PR #28, opened 2023-11-16
feat: enabling universal link

DylanRJ:
# iOS | Receive Auth Code and State from deeplink

This PR is to enable universal linking into the One Login app from domains for staging and build.
Much of the qualifying of the returned query parameters i.e. `authorizationCode` and `state` are handled as part of the [authentication package](https://github.com/govuk-one-login/mobile-ios-authentication)

# Checklist

### DIFF
--- a/Sources/Onboarding/MainCoordinator.swift
+++ b/Sources/Onboarding/MainCoordinator.swift
@@ -4,26 +4,30 @@ import Logging
 import UIKit
 
 final class MainCoordinator: NSObject,
+                             ParentCoordinator,
                              NavigationCoordinator {
-    private let window: UIWindow
     let root: UINavigationController
     let session: LoginSession
     let analyticsService: AnalyticsService
+    var childCoordinators = [ChildCoordinator]()
     private let viewControllerFactory = OnboardingViewControllerFactory.self
+    var authCoordinator: AuthenticationCoordinator?
     
-    init(window: UIWindow,
-         root: UINavigationController,
+    init(root: UINavigationController,
          session: LoginSession,
          analyticsService: AnalyticsService = OneLoginAnalyticsService()) {
-        self.window = window
         self.root = root
         self.session = session
         self.analyticsService = analyticsService
     }
     
     func start() {
+        authCoordinator = AuthenticationCoordinator(root: root,
+                                                    session: session)
         let introViewController = viewControllerFactory.createIntroViewController(analyticsService: analyticsService,
-                                                                                  session: session)
+                                                                                  session: session) {
+            self.openChildInline(self.authCoordinator!)
+        }
         root.setViewControllers([introViewController], animated: false)
     }
 }
--- a/Sources/Onboarding/MainCoordinator.swift
+++ b/Sources/Onboarding/MainCoordinator.swift
@@ -25,9 +25,9 @@ final class MainCoordinator: NSObject,
         authCoordinator = AuthenticationCoordinator(root: root,
                                                     session: session)
         let introViewController = viewControllerFactory.createIntroViewController(analyticsService: analyticsService,
-                                                                                  session: session) {
-            self.openChildInline(self.authCoordinator!)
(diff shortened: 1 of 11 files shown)


mcgarj (approved):

benmckillop (approved):

elishkaflint (approved):

Sources/Onboarding/MainCoordinator.swift
```diff
@@ -4,26 +4,26 @@ import Logging
...
-        self.window = window
         self.root = root
         self.session = session
         self.analyticsService = analyticsService
     }
     
     func start() {
-        let introViewController = viewControllerFactory.createIntroViewController(analyticsService: analyticsService,
-                                                                                  session: session)
+        let introViewController = viewControllerFactory.createIntroViewController(analyticsService: analyticsService) { [self] in
```
Oliver-Binns, line 24: It's not something for this pull request, but I've chatted with Robin on some thoughts around flow when opening the app: worth having a catch up with him around the way he's done similar things for GOV.UK.

I think our end-goal is:
- MainCoordinator which is a `TabBarCoordinator`.
- Tabs with child coordinators for Home, Wallet, etc.
- `MainCoordinator` opens the `AuthenticationCoordinator` flow modally (full-screen) on it's root (UITabBarController)

Tests/UnitTests/Onboarding/MainCoordinatorTests.swift
```diff
@@ -29,10 +28,20 @@ final class MainCoordinatorTests: XCTestCase {
...
-    func test_MainCoordinatorStart() throws {
+    func test_mainCoordinatorStart() throws {
         XCTAssertTrue(navigationController.viewControllers.count == 0)
         sut.start()
         XCTAssertTrue(navigationController.viewControllers.count == 1)
         XCTAssert(navigationController.topViewController is IntroViewController)
     }
+    
+    func test_mainCoordinatorOpensSubCoordinator() throws {
+        sut.start()
```
Oliver-Binns, line 39: Question for Elishka really:
Should we add any ACs in comments or anything within these tests to make it clear what they are doing?
This is only our first test and I'm not sure clear (from a user's perspective) on what it is testing at first glance.

Tests/UnitTests/Onboarding/MainCoordinatorTests.swift
```diff
@@ -29,10 +28,20 @@ final class MainCoordinatorTests: XCTestCase {
 }
 
 extension MainCoordinatorTests {
-    func test_MainCoordinatorStart() throws {
+    func test_mainCoordinatorStart() throws {
```
Oliver-Binns, line 31: `test_mainCoordinatorStart_displaysIntroViewController()` perhaps?

Sources/Onboarding/MainCoordinator.swift
DylanRJ: Interesting, will do!
My thoughts were very much down the line of a MainCoordinator which coordinates the other coordinators... 😂 
So essentially the MainCoordinator handles the Welcome screen - starts an AuthenticationCoordinator for Sign Up or Log In then finishes when that's done - then starts a TabBarCoordinator or something suchlike for the main function of the app.
Sounds a bit different to what you've mentioned above so yeah might be good for a wider conversation about that 🙂

mcgarj (approved):

Tests/UnitTests/Onboarding/MainCoordinatorTests.swift
elishkaflint: Yes that would be very helpful in general. On the back end, we typically do this in a test name with a convention along the lines of "when I do something, then it returns xyz". We could do that or comments - whatever people prefer in an iOS context.

DylanRJ: I can update all the tests to include comments in the format you've mentioned @elishkaflint 🙂

PR #80, opened 2024-03-04
feat: user sees 'unlock screen' when dismissing local auth OS prompt

mcgarj:
# DCMAW-8006: User sees 'unlock screen' when dismissing local auth OS prompt

This PR adds an unlock screen to the start of the user journey for returning users. If the user taps cancel on the Touch or Face ID prompt, the unlock button triggers the authentication again allowing the user to log in using local authentication

# Checklist


Sources/Onboarding/ViewModels/UnlockScreenViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+         primaryButtonAction: @escaping () -> Void ) {
+        self.analyticsService = analyticsService
+        self.primaryButtonViewModel = AnalyticsButtonViewModel(titleKey: "app_unlockButton",
+                                                               analyticsService: analyticsService) {
+            primaryButtonAction()
+        }
+    }
+
+    func didAppear() {
+        let screen = ScreenView(screen: IntroAnalyticsScreen.returningBiometricUser, titleKey: primaryButtonViewModel.title.stringKey)
```
DylanRJ, line 22: I'm not sure we want to be populating this `ScreenView`'s title with the button title, understand there isn't a title on this screen but we should probably check with analytics what they would want here in the long run. Potentially removing in the short term where we don't have that requirement yet 🙂

mcgarj: Happy to remove it if you think that's best, but it would anger sonar cloud and drop the coverage

DylanRJ: Very good point, maybe we just set a hard-coded screen value for this in the interim.
Just caught Sam (from analytics team) very quickly and he said that in the schema we'll be moving to, these title keys won't actually mean much so he's happy for us to put in a value for that 👍

Sources/Analytics/Onboarding/IntroAnalytics.swift
```diff
@@ -3,4 +3,5 @@ import Logging
 
 enum IntroAnalyticsScreen: String, LoggableScreen, NamedScreen {
     case welcomeScreen = "introWelcomeScreen"
+    case returningBiometricUser = "introReturningBiometricUser"
```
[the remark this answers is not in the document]
mcgarj: It's only used to pass something in for the `UnlockScreenViewModel`'s `didAppear` function 🙂

Sources/Onboarding/ViewModels/UnlockScreenViewModel.swift
mcgarj: Sounds good, so for `screen` remove the enum value and write any appropriate string for the time being?

DylanRJ (approved):

geoffreyssh97 (approved):

PR #87, opened 2024-03-20
feat: user sees sign in page when app fails to retrieve tokens from secure store

mcgarj:
# DCMAW-7977: User sees sign in page when app fails to retrieve tokens from secure store

This PR redirects a returning user to the signin page should a `SecureStoreError` be thrown when attempting to retrieve tokens

# Checklist

### DIFF
--- a/Sources/Login/LoginCoordinator.swift
+++ b/Sources/Login/LoginCoordinator.swift
@@ -59,7 +59,7 @@ final class LoginCoordinator: NSObject,
             tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
             finish()
         } catch is SecureStoreError {
-            try? userStore.secureStoreService.deleteItem(itemName: .accessTokenExpiry)
+            userStore.refreshStorage()
             start()
         } catch {
             print("Local Authentication error: \(error)")
--- a/Sources/Login/LoginCoordinator.swift
+++ b/Sources/Login/LoginCoordinator.swift
@@ -58,7 +58,7 @@ final class LoginCoordinator: NSObject,
         do {
             tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
             finish()
-        } catch is SecureStoreError {
+        } catch SecureStoreError.unableToRetrieveFromUserDefaults {
             userStore.refreshStorage()
             start()
         } catch {
--- a/Sources/Login/LoginCoordinator.swift
+++ b/Sources/Login/LoginCoordinator.swift
@@ -58,6 +58,9 @@ final class LoginCoordinator: NSObject,
         do {
             tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
             finish()
+        } catch is SecureStoreError {
+            try? userStore.secureStoreService.deleteItem(itemName: .accessTokenExpiry)
+            start()
         } catch {
             print("Local Authentication error: \(error)")
         }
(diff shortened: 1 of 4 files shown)


Sources/Login/LoginCoordinator.swift
```diff
@@ -58,6 +58,9 @@ final class LoginCoordinator: NSObject,
         do {
             tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
             finish()
+        } catch is SecureStoreError {
```
jdubey, line 61: The only kind of error readItem can throw is a SecureStoreError, so no need for two catch blocks.  They can be combined as one.  It will never be a "Local Authentication error", that labeling is incorrect.

jdubey (approved):

Tests/UnitTests/Login/LoginCoordinatorTests.swift
```diff
@@ -152,11 +149,14 @@ extension LoginCoordinatorTests {
...
         // GIVEN the secure store returns an error from reading an item
-        mockSecureStore.errorFromReadItem = SecureStoreError.generic
+        mockSecureStore.errorFromReadItem = SecureStoreError.unableToRetrieveFromUserDefaults
         // WHEN the LoginCoordinator's getAccessToken method is called
         sut.getAccessToken()
         // THEN the token holder's access token property should not get the access token from secure store
         XCTAssertEqual(sut.tokenHolder.accessToken, nil)
+        // THEN login flow should be triggered
+        XCTAssertTrue(sut.root.viewControllers.count == 1)
+        XCTAssertTrue(sut.root.topViewController is IntroViewController)
```
DylanRJ, line 159: I think you could probably add something to the mocks to validate the `refreshStorage()` method is being called 🙂

Sources/Login/LoginCoordinator.swift
DylanRJ: Agree this should be a catch all I think, unless cancelling the local auth prompts throws an error..? I don't think I've tested that... we wouldn't want any action in that case, would just want to return and the user to be able to try again

Tests/UnitTests/Mocks/Sessions+Services/Storage/MockDefaultsStore.swift
```diff
@@ -13,6 +14,7 @@ class MockDefaultsStore: DefaultsStorable {
     }
     
     func removeObject(forKey defaultName: String) {
+        self.didCallRemoveObject = true
```
DylanRJ, line 17: I imagine you could probably just remove the object set in the test: `mockDefaultStore.set(Date() + 60, forKey: .accessTokenExpiry)` and assert that that is nil?

mcgarj: Would it be better to check that all of `mockDefaultStore`'s `savedData` is empty or to specifically assert that the access token expiry is nil?

Sources/Login/LoginCoordinator.swift
mcgarj: With some testing of this, I've had to change the first catch to `catch SecureStoreError.unableToRetrieveFromUserDefaults` and then have a general catch all below. With a general catch all, or for any `SecureStoreError` a returning user on the unlock screen cancelling Touch ID would be sent to the sign in page

mcgarj: Do we need to handle `cantDecryptData` thrown by `decryptDataWithPrivateKey` in the same catch block as `unableToRetrieveFromUserDefaults`?

DylanRJ (approved):

PR #103, opened 2024-04-16
feat: implement GA4 schema on the Unlock page

mcgarj:
# DCMAW-8394: Implement GA4 schema on the Unlock page

This adds the GA4 schema to the unlock screen and `trackEvent` for the unlock button

# Checklist

- [ ] ~~Met all accessibility requirements?~~
    - [ ] ~~Checked dynamic type sizes are applied~~
    - [ ] ~~Checked VoiceOver can navigate your new code~~
    - [ ] ~~Checked a user can navigate only using a keyboard around your new code~~

### DIFF
--- a/Sources/Analytics/Onboarding/BiometricEnrollmentAnalyticsScreen.swift
+++ b/Sources/Analytics/Onboarding/BiometricEnrollmentAnalyticsScreen.swift
@@ -10,4 +10,5 @@ enum BiometricEnrollmentAnalyticsScreen: String, ScreenType {
 enum BiometricEnrollmentAnalyticsScreenID: String {
     case faceIDEnrollment = "7590abba-f48b-4790-be64-f6ffc3dd35e2"
     case touchIDEnrollment = "cc1db40a-2a5a-43fe-a1fe-fc85b55fdcab"
+    case unlockScreen = "d61fcd23-cbcf-4df6-8e9c-d0dec956dbe1"
 }
--- a/Sources/Application/SceneLifecycle.swift
+++ b/Sources/Application/SceneLifecycle.swift
@@ -1,4 +1,5 @@
 import GAnalytics
+import GDSAnalytics
 import Logging
 import UIKit
 
@@ -18,6 +19,10 @@ extension SceneLifecycle {
     
     func promptToUnlock() {
         coordinator?.evaluateRevisit {
+            let screen = ScreenView(id: BiometricEnrollmentAnalyticsScreenID.unlockScreen.rawValue,
+                                    screen: BiometricEnrollmentAnalyticsScreen.unlockScreen,
+                                    titleKey: "one login unlock screen")
+            analyticsService.trackScreen(screen)
             windowManager?.hideUnlockWindow()
         }
     }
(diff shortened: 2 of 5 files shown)


jdubey (commented):
So as I mentioned in Slack, this doesn't quite work because events won't send when the app is in the background, and that is when didAppear gets called at the moment.   This is evidenced in the console logs:

10.19.0 - [FirebaseAnalytics][I-ACS023142] Cancelling background upload task. Task ID : 144
10.19.0 - [FirebaseAnalytics][I-ACS031026] Cannot log screen view event when the app is not active

I've played with it a little and I think this can be solved by implementing 

`func sceneWillResignActive(_ scene: UIScene)`

in SceneDelegate, and calling `displayUnlockScreen()` there instead of in `sceneDidEnterBackground`

DylanRJ:
> So as I mentioned in Slack, this doesn't quite work because events won't send when the app is in the background, and that is when didAppear gets called at the moment. This is evidenced in the console logs:
> 
> 10.19.0 - [FirebaseAnalytics][I-ACS023142] Cancelling background upload task. Task ID : 144 10.19.0 - [FirebaseAnalytics][I-ACS031026] Cannot log screen view event when the app is not active
> 
> I've played with it a little and I think this can be solved by implementing
> 
> `func sceneWillResignActive(_ scene: UIScene)`
> 
> in SceneDelegate, and calling `displayUnlockScreen()` there instead of in `sceneDidEnterBackground`

Ah, okay, didn't see that chat in Slack.
Wondering if it 'holds on' to that event and waits for the app to come back to the foreground to upload it... it seems to do that when a network connection isn't present but yeah could be different with this.
In terms of using `func sceneWillResignActive(_ scene: UIScene)`, personally I think that's fine but will mark a slightly different experience as the lock screen will display as soon as the user interacts with the device navigation, akin to a banking app. I think it would probably be worth mentioning that to product/design.

jdubey:
> > So as I mentioned in Slack, this doesn't quite work because events won't send when the app is in the background, and that is when didAppear gets called at the moment. This is evidenced in the console logs:
> > 10.19.0 - [FirebaseAnalytics][I-ACS023142] Cancelling background upload task. Task ID : 144 10.19.0 - [FirebaseAnalytics][I-ACS031026] Cannot log screen view event when the app is not active
> > I've played with it a little and I think this can be solved by implementing
> > `func sceneWillResignActive(_ scene: UIScene)`
> > in SceneDelegate, and calling `displayUnlockScreen()` there instead of in `sceneDidEnterBackground`
> 
> Ah, okay, didn't see that chat in Slack. Wondering if it 'holds on' to that event and waits for the app to come back to the foreground to upload it... it seems to do that when a network connection isn't present but yeah could be different with this. In terms of using `func sceneWillResignActive(_ scene: UIScene)`, personally I think that's fine but will mark a slightly different experience as the lock screen will display as soon as the user interacts with the device navigation, akin to a banking app. I think it would probably be worth mentioning that to product/design.

I did check for delayed upload, waiting around for 5 minutes or so per log message of 

10.19.0 - [FirebaseAnalytics][I-ACS002002] Engagement timer scheduled to fire in approx. (s): 3600

Though now that I redo the math, I should have waited 6 minutes :-P.  However, based on other log messages referenced above, and the fact that the more logical behavior would be for it to fire any queued events as soon as the app is active, it feels like it just gives up

PR #108, opened 2024-04-18
fix: dcmaw-8877 encryption keys not re created if user has passcode but no biometrics

jdubey:
# DCMAW-8877 iOS | Encryption keys not re-created if user has passcode but no biometrics set up

Add ability to test whether device has a passcode setup but is not enrolled in biometrics.  Use this test to configure the access control of the SecureStorage appropriately.
Wrap keychain access in a Task on the main thread.  This is necessary because on some devices (like my iPad), without this change the passcode entry keypad appears "behind" the splash or unlock page, and will not display until the app is minimized, making it impossible to unlock.

# Checklist

### DIFF
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -44,25 +44,21 @@ final class MainCoordinator: NSObject,
         loginCoordinator?.handleUniversalLink(url)
     }
     
-    func evaluateRevisit(action: @escaping () -> Void) {
-        Task {
-            await MainActor.run {
-                if userStore.returningAuthenticatedUser {
-                    do {
-                        tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
-                        homeCoordinator?.updateToken(accessToken: tokenHolder.accessToken)
-                        action()
-                    } catch {
-                        print("Error getting token: \(error)")
-                    }
-                } else if tokenHolder.validAccessToken || tokenHolder.accessToken == nil {
-                    action()
-                } else {
-                    tokenHolder.accessToken = nil
-                    start()
-                    action()
-                }
+    func evaluateRevisit(action: () -> Void) {
+        if userStore.returningAuthenticatedUser {
+            do {
+                tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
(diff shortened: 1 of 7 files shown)


Sources/Application/SceneLifecycle.swift
```diff
@@ -19,7 +19,7 @@ extension SceneLifecycle {
     
     func promptToUnlock() {
         coordinator?.evaluateRevisit {
-            windowManager?.hideUnlockWindow()
+            self.windowManager?.hideUnlockWindow()
```
DylanRJ, line 22: Is this needed?

jdubey: Yes, because this is now an escaping closure.  Compiler complains otherwise

DylanRJ: Worth capturing self as unowned?
```swift
func promptToUnlock() {
    coordinator?.evaluateRevisit { [unowned self] in
        windowManager?.hideUnlockWindow()
    }
}
```
?

Sources/Application/MainCoordinator.swift
```diff
@@ -44,21 +44,25 @@ final class MainCoordinator: NSObject,
...
-        if userStore.returningAuthenticatedUser {
-            do {
-                tokenHolder.accessToken = try userStore.secureStoreService.readItem(itemName: .accessToken)
-                homeCoordinator?.updateToken(accessToken: tokenHolder.accessToken)
-                action()
-            } catch {
-                print("Error getting token: \(error)")
+    func evaluateRevisit(action: @escaping () -> Void) {
+        Task {
+            await MainActor.run {
```
DylanRJ, line 49: Wouldn't marking this method (and the other dealing with local auth in the `LoginCoordinator`) as `@MainActor` achieve essentially the same thing without having to include a Task and having the closure be `@escaping`?
There was a pretty sizeable refactor to do away with using `waitForTruth` where it isn't absolutely necessary and avoid flakey tests so would certainly be on the side of that where possible.

jdubey: Still doesn't work.  I admit this is an edge case, as it only seems to happen on my iPad, so it is unlikely to be encountered by other users.  On the other hand, you are pretty much blocked from using the app with local auth if you run into it.

mcgarj (approved):

PR #112, opened 2024-04-22
feat: dcmaw-8571 privacy policy link button

jdubey:
# DCMAW-8571: iOS | Add privacy policy link to analytics opt-in page

Add the privacy policy link to the Analytics opt-in page.  This leverages changes made in mobile-ios-common

# Checklist


OneLogin.xcodeproj/project.pbxproj
```diff
@@ -2055,11 +2060,11 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = N8W395F695;
+				EXTERNAL_BASE_URL = signin.account.gov.uk;
 				FEATURE_FLAG_FILE = FeatureFlagsBuild;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = "OneLogin-Info.plist";
-				INFOPLIST_KEY_CFBundleDisplayName = "One Login - Build";
-				INFOPLIST_KEY_NSCameraUsageDescription = "This app does not use the camera";
```
DylanRJ, line 2062: Looks like the Xcode project file is doing some annoying stuff here in terms of renaming the app display name and getting rid of the camera usage description. I've seen this happen locally as well, very annoying.
Worth keeping an eye out for changes in this project file, we've fallen foul of stuff like this on the program in the past with stuff making it's way into production... less of a concern for this app atm but good practice to get into 🙂
With bringing these project settings back, there's a nice way of toggling settings names/titles in Xcode (if you don't know about it already).
When you're in the project file, Editor -> Show Setting Names will show them in the format you see here and may make it a little easier
image 2024-04-22 at 14 31 56

OneLogin.xcworkspace/xcshareddata/swiftpm/Package.resolved
DylanRJ, line 1: I think we should be careful about updating these transitive dependencies where they haven't come from updating our top-level packages. I've not heard a definitive answer on this but ideally we need our packages' `Package.resolved` files to match this file

jdubey: So if I understand what you are saying, every package that is dependent on mobile-ios-common should get updated to this new revision?

OneLogin.xcodeproj/project.pbxproj
mcgarj: Is the only way to change these to revert them manually to what they are in the commit changes? And there's no way to stop Xcode randomly doing this?

DylanRJ (approved):

mcgarj (approved):

OneLogin.xcworkspace/xcshareddata/swiftpm/Package.resolved
DylanRJ: Hopefully dependabot will resolve these versions to what the top-level package dependencies have in their `Package.resolved` files.
If not, we probably need to think about how we can manage this going forward

PR #121, opened 2024-04-30
feat: authenticated request to Hello World API

mcgarj:
# DCMAW-8805: Authenticated request to Hello World API

This PR sends an authenticated request via STS to the Hello World API

# Checklist


DylanRJ (approved):
🙌

Sources/Screens/DeveloperMenu/DeveloperMenuViewController.swift
```diff
@@ -1,16 +1,121 @@
...
+    }
+    
+    @IBOutlet private var unhappyPathResultLabel: UILabel! {
+        didSet {
+            unhappyPathResultLabel.isHidden = true
+            unhappyPathResultLabel.accessibilityIdentifier = "sts-unhappy-path-result"
+        }
+    }
+
+    private func formatResultLabel(label: UILabel,
```
jdubey, line 111: Because passing an instance variable into a function always feels wrong to me, I might suggest this solution instead:

```
fileprivate extension UILabel {
    func showErrorMessage(_ error: ServerError? = nil) {
        textColor = .red
        isHidden = false
        if let error {
            self.text = "Error code: \(error.errorCode)\nEndpoint: \(error.endpoint ?? "missing")"
        } else {
            self.text = "Error"
        }
    }
    
    func showSuccessMessage(_ message: String) {
        self.textColor = .gdsGreen
        isHidden = false
        text = message
    }
}
```
Note I am not setting font here.  I would do that in the didSet of the labels themselves, as I don't see a reason we would change this based on response.

Sources/Utilities/Network/NetworkClientele.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+import Networking
+
+protocol NetworkClientele {
```
jdubey, line 4: I feel like this can have a better name.  RequestAuthorizing?  AuthorizedRequester?

jdubey (approved):

Sources/Screens/DeveloperMenu/DeveloperMenuViewController.swift
mcgarj: Good shout, thanks! I've not used fileprivate much, but in this case is it best to include this in the ViewController or make a file as an extension of UILabel (like what we have for String and LAContext)?

Sources/Screens/DeveloperMenu/DeveloperMenuViewController.swift
```diff
@@ -1,16 +1,125 @@
...
+    }
+    
+    private func helloWorldHappyPath() {
+        Task {
+            do {
+                let data = try await networkClient?.makeAuthorizedRequest(exchangeRequest: URLRequest(url: AppEnvironment.stsToken),
+                                                                          scope: "sts-test.hello-world.read",
+                                                                          request: URLRequest(url: AppEnvironment.stsHelloWorld))
+                happyPathResultLabel.showSuccessMessage("Success: \(String(data: data!, encoding: .utf8) ?? "no body")")
+//                formatResultLabel(label: happyPathResultLabel,
```
jdubey, line 48: Please remove commented code

jdubey (approved):

DylanRJ (approved):

PR #129, opened 2024-05-13
feat: dcmaw-8570 verify and store the id token

jdubey:
# DCMAW-8570: 

This feature utilizes [jwt-kit](https://github.com/vapor/jwt-kit) to decode and verify the idToken when returned from STS /token response

The token is verified using a JavaWebKey (JWK).  JWKs are fetched from a new endpoint, which is included as part of this feature.

After the token is verified, it is stored in the Secure Store.  Should any errors occur, the user is shown the login error page

# Checklist

- [ ] Created a `draft` pull request if it is not yet ready for review~~


Tests/UnitTests/Login/AuthenticationCoordinatorTests.swift
```diff
@@ -170,4 +170,16 @@ extension AuthenticationCoordinatorTests {
...
+        mockLoginSession.errorFromPerformLoginFlow = JWTVerifierError.unableToFetchJWKs
+        sut.start()
+        // GIVEN the AuthenticationCoordinator has logged in via start()
+        // WHEN the AuthenticationCoordinator calls performLoginFlow on the session
+        // and user cancelled the login modal
+        waitForTruth(self.mockLoginSession.didCallPerformLoginFlow, timeout: 20)
+        // THEN user is returned to the intro screen
+        // THEN the loginError should be a userCancelled error
+        sut.loginError = JWTVerifierError.unableToFetchJWKs
+    }
```
DylanRJ, line 184: Might be nice to assert that the right error screen is being shown for this one, not totally necessary but shouldn't be too difficult.
Also comments still pertain to the userCancelled error here 👍

DylanRJ:
Do we think we should standardise the approach and move the `Staging` build configuration to `Debug_Staging` like we're now doing with build. Would unit tests run on the Staging scheme not fail without this?

mcgarj (approved):

jdubey:
> Do we think we should standardise the approach and move the `Staging` build configuration to `Debug_Staging` like we're now doing with build. Would unit tests run on the Staging scheme not fail without this?

If we add tests to the Staging scheme that have an @testable import, then yes, we would need to do this. But otherwise the way tests are set up now, it appears to work fine.

PR #135, opened 2024-05-20
feat: disable/enable sign in buttton

DylanRJ:
# DCMAW-8928: iOS | Disable and add loading icon to Sign In button

Previously the "Sign in" button on the intro screen was not disabled when interacted with and could produce incorrect behaviour if the user interacted with the button again while the authentication modal was launching.
This PR implements disabling the button and reenabling it when the authentication modal is dismissed.

# Checklist

### DIFF
--- a/Sources/Login/AuthenticationCoordinator.swift
+++ b/Sources/Login/AuthenticationCoordinator.swift
@@ -7,7 +7,7 @@ final class AuthenticationCoordinator: NSObject,
                                        ChildCoordinator,
                                        NavigationCoordinator {
     let root: UINavigationController
-    var parentCoordinator: ParentCoordinator?
+    weak var parentCoordinator: ParentCoordinator?
     let session: LoginSession
     let analyticsService: AnalyticsService
     let errorPresenter = ErrorPresenter.self
--- a/Sources/Login/AuthenticationCoordinator.swift
+++ b/Sources/Login/AuthenticationCoordinator.swift
@@ -67,6 +67,9 @@ final class AuthenticationCoordinator: NSObject,
     
     func handleUniversalLink(_ url: URL) {
         do {
+            if let loginCoordinator = parentCoordinator as? LoginCoordinator {
+                loginCoordinator.introViewController?.enableIntroButton()
+            }
             try session.finalise(redirectURL: url)
         } catch {
             let genericErrorScreen = errorPresenter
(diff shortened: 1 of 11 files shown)


DylanRJ:
@jdubey @mcgarj
Going to leave some comments on the implementation here as it's a bit strange

Sources/Login/AuthenticationCoordinator.swift
```diff
@@ -67,6 +67,9 @@ final class AuthenticationCoordinator: NSObject,
     
     func handleUniversalLink(_ url: URL) {
         do {
+            if let loginCoordinator = parentCoordinator as? LoginCoordinator {
+                loginCoordinator.introViewController?.enableIntroButton()
+            }
```
DylanRJ, line 72: I've made `LoginCoordinator.introViewController` an internal property so that we can call it here in the `AuthenticationCoordinator`. I did this rather than passing it in to make it more simplistic, happy to change if there is a better way to do this.
This is done here because when selecting a universal link, which is fed through to this method, the below call to `AppAuthSession.finalise(edirectURL:)` immediately dismisses the authentication modal before triggering the above `AppAuthSession.performLoginFlow(configuration:)`. This means that the button loading will remain until `finish()` is called on line 36. This isn't terrible UX and could be fine where we're going; presenting a loading screen on top of this screen until the work to call the `/token` endpoint is complete and we finish this coordinator successfully or with an error. Though, for the meantime, I have implemented this call here to ensure the button is reenabled.

Sources/Login/LoginCoordinator.swift
```diff
@@ -81,6 +83,7 @@ final class LoginCoordinator: NSObject,
                 } else {
                     let networkErrorScreen = errorPresenter
                         .createNetworkConnectionError(analyticsService: analyticsCenter.analyticsService) { [unowned self] in
+                            introViewController?.enableIntroButton()
```
DylanRJ, line 86: Because we haven't launched the `AuthenticationCoordinator` at this point, we need to ensure this is called on the network error button action.

Sources/Login/LoginCoordinator.swift
```diff
@@ -128,6 +132,7 @@ extension LoginCoordinator: ParentCoordinator {
         case _ as OnboardingCoordinator:
             return
         case let child as AuthenticationCoordinator where child.loginError != nil:
+            introViewController?.enableIntroButton()
```
DylanRJ, line 135: When the user hits cancel on the authentication modal, `AuthenticationCoordinator.handleUniversalLink(_ URL:)` isn't called but an error is logged and the coordinator finishes so we need to ensure the button is reenabled at this point.
This method will be called twice when the `AuthenticationCoordinator` finishes with an error after `handleUniversalLink(_ URL:)` is called but that should be fine.

jdubey (approved):

mcgarj (approved):

PR #136, opened 2024-05-28
feat:  dcmaw-8544 user can see email on home page

jdubey:
# DCMAW-8544 | User can see their email on the home page

After successful login, decode the idToken to extract the e-mail, and display on the home page

Added additional functionality to JWTVerifier to decode the idToken without performing verification, since at the point of decoding, the idToken will already have been verified, whether it has been freshly update from BE or retrieved from secure store (as we wouldn't store an unverified token).  This is to prevent the need from having to make another JWKS call every time we decode.

Also refactored so that the initial biometrics/passcode request is for the idToken, not the access token.   The access token is retrieved from the secure store when the Developer Menu is visited, as that is when it is actually required.  This will invoke another biometric auth request, as that is the nature of how secure store works, i.e., every access to the store requires retrieval of the private key from the secure enclave, and every secure enclave access requires biometric authentication.

# Checklist

~~- [ ] Created a `draft` pull request if it is not yet ready for review~~

### DIFF
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -76,8 +76,8 @@ final class MainCoordinator: NSObject,
                                   analyticsCenter: analyticsCenter,
                                   networkMonitor: NetworkMonitor.shared,
                                   userStore: userStore,
-                                  tokenHolder: tokenHolder,
-                                  startupError: error)
+                                  tokenHolder: tokenHolder)
+        lc.startupError = error
         openChildModally(lc, animated: false)
         loginCoordinator = lc
     }
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -14,9 +14,9 @@ final class MainCoordinator: NSObject,
     let userStore: UserStorable
     let tokenHolder = TokenHolder()
     var networkClient: NetworkClient?
-    private var loginCoordinator: LoginCoordinator?
-    private var homeCoordinator: HomeCoordinator?
-    private var profileCoordinator: ProfileCoordinator?
+    private weak var loginCoordinator: LoginCoordinator?
+    private weak var homeCoordinator: HomeCoordinator?
+    private weak var profileCoordinator: ProfileCoordinator?
     
     init(windowManager: WindowManagement,
          root: UITabBarController,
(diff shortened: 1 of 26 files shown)


mcgarj (approved):

Sources/Application/MainCoordinator.swift
```diff
@@ -45,12 +46,18 @@ final class MainCoordinator: NSObject,
...
+                            throw SecureStoreError.unableToRetrieveFromUserDefaults
+                        }
+                        tokenHolder.idTokenPayload = try tokenVerifier.extractPayload(idToken)
                         updateToken()
                         action()
                     } catch {
-                        print("Error getting token: \(error)")
+                        loginCoordinator?.root.dismiss(animated: false)
+                        tokenHolder.accessToken = nil
+                        userStore.defaultsStore.set(nil, forKey: .accessTokenExpiry)
```
DylanRJ, line 58: Would calling `UserStorable.clearTokenInfo()` make sense here?

Sources/Login/LoginCoordinator.swift
```diff
@@ -92,7 +95,16 @@ final class LoginCoordinator: NSObject,
                     root.pushViewController(networkErrorScreen, animated: true)
                 }
             }
+
         root.setViewControllers([rootViewController], animated: true)
+        if let startupError {
```
DylanRJ, line 100: Wondering if we could give this a more descriptive name just so it's clear they user has been unable to either get the id token or verify it?

DylanRJ (approved):
There's a fair amount of complexity with the id token reading, verification and extraction. Might be useful to do an internal team recce at some point to make sure we're aligned 👍

Sources/Login/LoginCoordinator.swift
```diff
@@ -49,27 +53,26 @@ final class LoginCoordinator: NSObject,
...
+                        throw SecureStoreError.unableToRetrieveFromUserDefaults
+                    }
+                    tokenHolder.idTokenPayload = try tokenVerifier.extractPayload(idToken)
                     windowManager.hideUnlockWindow()
                     root.dismiss(animated: false)
                     finish()
-                } catch SecureStoreError.unableToRetrieveFromUserDefaults,
-                        SecureStoreError.cantInitialiseData,
-                        SecureStoreError.cantRetrieveKey {
+                } catch {
```
DylanRJ, line 72: Would we not want to add the error to the `startupError` property here?

PR #155, opened 2024-06-23
refactor: moving local auth code out of LoginCoordinator

DylanRJ:
# refactor: moving local auth code out of LoginCoordinator

_Thank you for your contribution to the project._

_Please enter a short description of your change here._

# Checklist

~- [ ] Met all of the acceptance criteria specified in the user story on Jira~

### DIFF
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -108,7 +108,7 @@ extension MainCoordinator {
                                   userStore: userStore,
                                   networkMonitor: NetworkMonitor.shared,
                                   tokenHolder: tokenHolder)
-        if error is AuthError {
+        if !(error is AuthError) {
             lc.loginError = error
         }
         openChildModally(lc, animated: false)
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -35,7 +35,53 @@ final class MainCoordinator: NSObject,
     func start() {
         root.delegate = self
         addTabs()
-        showLogin()
+        if userStore.validAuthenticatedUser {
+            windowManager.displayUnlockWindow(analyticsService: analyticsCenter.analyticsService) { [unowned self] in
+                evaluateRevisit()
+            }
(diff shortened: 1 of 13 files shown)


Sources/Application/MainCoordinator.swift
```diff
@@ -35,81 +35,80 @@ final class MainCoordinator: NSObject,
...
-    
-    func handleUniversalLink(_ url: URL) {
-        switch UniversalLinkQualifier.qualifyOneLoginUniversalLink(url) {
-        case .login:
-            loginCoordinator?.handleUniversalLink(url)
-        case .wallet:
-            walletCoordinator?.walletSDK.deeplink(with: url.absoluteString)
-        case .unknown:
-            return
+        if userStore.validAuthenticatedUser {
```
jdubey, line 38: I've been playing with this a little, and wonder if we need this check, as evaluateRevisit performs the same check and also does full login if the check does not pass.

The only difference is unlock window will be displayed on start, until fullLogin is called.  But since the unlock window doesn't differ from the splash screen, there is no perceptible change in behavior I could see, as for some reason, the "Unlock" button does not appear.  This is even if I set a breakpoint and step through the start up process.   And the code execution through to fullLogin() all happens almost instantaneously anyway.
So start could just be
```
    func start() {
        root.delegate = self
        addTabs()
        windowManager.displayUnlockWindow(analyticsService: analyticsCenter.analyticsService) { [unowned self] in
            evaluateRevisit()
        }
        evaluateRevisit()
    }
```
which is a bit cleaner.   Feel free to ignore if I'm missing something :-)

jdubey (approved):

Sources/Application/MainCoordinator.swift
DylanRJ: Yes, good shout. I was thinking of how I could do the inverse; remove the `if userStore.validAuthenticatedUser` from `evaluateRevisit()` but yeah that sounds good!

gabby-san-digco (approved):

PR #156, opened 2024-06-25
feat: delete wallet credential on sign out

DylanRJ:
# DCMAW-9184: iOS | Delete Wallet credentials when the user Signs out

This PR implements wallet credential deletion when the user selects to sign out of the app. The token information and analytics preferences are also removed.

# Checklist

### DIFF
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -172,9 +172,19 @@ extension MainCoordinator: ParentCoordinator {
         case _ as LoginCoordinator:
             updateToken()
         case _ as ProfileCoordinator:
-            fullLogin()
-            homeCoordinator?.baseVc?.isLoggedIn(false)
-            root.selectedIndex = 0
+            do {
+                try walletCoordinator?.clearWallet()
+                fullLogin()
+                homeCoordinator?.baseVc?.isLoggedIn(false)
+                root.selectedIndex = 0
+            } catch {
+                let errorVC = ErrorPresenter.createSignoutError(errorDescription: error.localizedDescription,
+                                                                analyticsService: analyticsCenter.analyticsService) {
+                    exit(0)
+                }
+                errorVC.modalPresentationStyle = .overFullScreen
+                root.present(errorVC, animated: false)
+            }
         default:
             break
         }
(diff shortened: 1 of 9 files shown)


Sources/Application/MainCoordinator.swift
```diff
@@ -171,12 +171,33 @@ extension MainCoordinator: ParentCoordinator {
...
-        case _ as ProfileCoordinator:
-            fullLogin()
-            homeCoordinator?.baseVc?.isLoggedIn(false)
-            root.selectedIndex = 0
         default:
             break
         }
     }
+    
+    func performChildCleanup(child: ChildCoordinator) {
```
mcgarj, line 179: More of a question than anything: has the functionality of this method been put here so we can switch on any future children and make life a bit easier? Or are there also performance/functional reasons behind moving most of the functionality from ProfileCoordinator?

mcgarj (approved):

jdubey (approved):
Looks good.  One favour to ask

Please update in the `create_keychain` fastfile like so

```
      create_keychain(
        name: "signing_temp",
        password: password,
        timeout: 1800,
        unlock: true
      )

```
This increases the timeout value before the keychain locks again to 30 minutes.  This is what was causing the signing of Realm to hang during deployment, as the default is 5 minutes.  30 minutes might be overkill but I'd rather the builds be reliable.  We can bring it down after we get a sense of new build times.

Sources/Application/MainCoordinator.swift
DylanRJ: So this `performChildCleanup` method is part of the [Coordination package](https://github.com/govuk-one-login/mobile-ios-coordination/blob/main/Sources/Coordination/ParentCoordinator.swift) and is usually called as part of `childDidFinish` which is actually what `finish()` calls from the child coordinator.
childDidFinish removes the coordinator from the list of child coordinators on the parent though which isn't really what we want for our tab coordinators, so being able to call this method allows us to react to behaviour from the child on the parent. It's maybe a bit hacky but a useful way of us being able to reuse code from coordination.
As you say, we can switch on children in future if we want to do this from more tab coordinators to main or other child coordinators for any other parent coordinators 🙂

PR #160, opened 2024-07-04
feat: reauth on expired accessToken

DylanRJ:
# DCMAW-9511: iOS | User sees 'You have been signed out' re-auth page when user's access token expires (whilst in-app)

When a user’s access token has expired and they call the hello-world API, they will need to re-authenticate using the One Login WebView in order to get a new access token. Once the user has successfully re-authenticated, they will be returned to their current place in-app.
Additionally, when a user returns to the app with an expired access token, the same behaviour should occur but infront of the intro screen.

# Checklist

### DIFF
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -12,7 +12,7 @@ final class MainCoordinator: NSObject,
     var childCoordinators = [ChildCoordinator]()
     var analyticsCenter: AnalyticsCentral
     let userStore: UserStorable
-    let tokenHolder = TokenHolder()
+    let tokenHolder = TokenHolder.shared
     private var tokenVerifier: TokenVerifier
     
     private weak var loginCoordinator: LoginCoordinator?
@@ -99,8 +99,7 @@ extension MainCoordinator {
                                   root: UINavigationController(),
                                   analyticsCenter: analyticsCenter,
                                   userStore: userStore,
-                                  networkMonitor: NetworkMonitor.shared,
-                                  tokenHolder: tokenHolder)
+                                  networkMonitor: NetworkMonitor.shared)
         lc.loginError = error
         openChildModally(lc, animated: false)
         loginCoordinator = lc
@@ -113,9 +112,9 @@ extension MainCoordinator {
     }
     
     private func addHomeTab() {
-        let hc = HomeCoordinator(analyticsService: analyticsCenter.analyticsService,
-                                 userStore: userStore,
-                                 tokenHolder: tokenHolder)
+        let hc = HomeCoordinator(window: windowManager.appWindow,
+                                 analyticsService: analyticsCenter.analyticsService,
+                                 userStore: userStore)
         addTab(hc)
         homeCoordinator = hc
     }
@@ -123,8 +122,7 @@ extension MainCoordinator {
     private func addWalletTab() {
         let wc = WalletCoordinator(window: windowManager.appWindow,
                                    analyticsService: analyticsCenter.analyticsService,
-                                   secureStoreService: userStore.authenticatedStore,
-                                   tokenHolder: tokenHolder)
+                                   secureStoreService: userStore.authenticatedStore)
         addTab(wc)
         walletCoordinator = wc
     }
@@ -132,7 +130,6 @@ extension MainCoordinator {
     private func addProfileTab() {
         let pc = ProfileCoordinator(analyticsService: analyticsCenter.analyticsService,
                                     userStore: userStore,
-                                    tokenHolder: tokenHolder,
                                     urlOpener: UIApplication.shared)
         addTab(pc)
         profileCoordinator = pc
--- a/Sources/Application/MainCoordinator.swift
+++ b/Sources/Application/MainCoordinator.swift
@@ -58,7 +58,7 @@ final class MainCoordinator: NSObject,
                 }
             }
         } else {
-            fullLogin()
+            fullLogin(error: TokenError.expired)
         }
     }
     
@@ -68,13 +68,13 @@ final class MainCoordinator: NSObject,
             SecureStoreError.unableToRetrieveFromUserDefaults,
             SecureStoreError.cantInitialiseData,
             SecureStoreError.cantRetrieveKey:
-            fullLogin(error)
+            fullLogin(error: error)
         default:
             print("Token retrival error: \(error)")
         }
     }
     
-    private func fullLogin(_ error: Error? = nil) {
+    private func fullLogin(error: Error? = nil) {
         tokenHolder.clearTokenHolder()
         userStore.refreshStorage(accessControlLevel: LAContext().isPasscodeOnly ? .anyBiometricsOrPasscode : .currentBiometricsOrPasscode)
         showLogin(error)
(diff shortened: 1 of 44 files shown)


DylanRJ:
This is a pretty large PR so going to leave some code comments to highlight the most critical parts of the implementation.

Sources/Application/MainCoordinator.swift
```diff
@@ -12,8 +12,9 @@ final class MainCoordinator: NSObject,
     var childCoordinators = [ChildCoordinator]()
     var analyticsCenter: AnalyticsCentral
     let userStore: UserStorable
-    let tokenHolder = TokenHolder()
+    let tokenHolder = TokenHolder.shared
     private var tokenVerifier: TokenVerifier
+    static var reauth = false
```
DylanRJ, line 17: I spent some time thinking/researching how this could be done for reauth scenarios where the user is already within the app (LoginCoordinator is not being presented).
I had a look into ObservableObject and posting notifications but this felt like the easiest and cleanest way to do it. Would be happy to change this if there's a more preferable means of achieving this.

Sources/Application/MainCoordinator.swift
```diff
@@ -42,19 +43,23 @@ final class MainCoordinator: NSObject,
...
+                            tokenHolder.idTokenPayload = try tokenVerifier.extractPayload(idToken)
+                            updateToken()
+                            windowManager.hideUnlockWindow()
+                        } catch {
+                            handleLoginError(error)
+                        }
                     }
                 }
+            } else {
+                fullLogin(error: TokenError.expired)
```
DylanRJ, line 62: This flow is now driven from two points in the `userStore`:
1. If the user has any accessTokenExpiry date stored in `UserDefaults` at all - telling us whether we have any record of them previously logging in before.
2. If the user has an expired accessTokenExpiry date stored in `UserDefaults` - telling us that they have logged in before but their token is now expired and they need to do a full log in to use the app.

This allows us to display a sign out warning screen over the intro screen if we can see that they have previously authenticated, but their access token has now expired. Rather than someone who has not authenticated before and thus no sign out warning screen is needed.

Sources/Application/TokenHolder.swift
```diff
@@ -3,16 +3,23 @@ import Networking
 
 enum TokenError: Error {
     case bearerNotPresent
+    case expired
 }
 
 class TokenHolder: AuthenticationProvider {
+    static let shared = TokenHolder()
```
DylanRJ, line 10: I've made a singleton for the TokenHolder as there should only ever be one place where we hold tokens in memory for the app. This avoids having to pass it around everywhere and makes it easily accessible from anywhere it's needed. Happy to revert if we think this is insecure or prone to other issues.

Sources/Tabs/HomeCoordinator.swift
```diff
@@ -43,16 +48,39 @@ final class HomeCoordinator: NSObject,
...
+            MainCoordinator.reauth = true
+            root.dismiss(animated: true)
+            tokenHolder.clearTokenHolder()
+            userStore.refreshStorage(accessControlLevel: LAContext().isPasscodeOnly ? .anyBiometricsOrPasscode : .currentBiometricsOrPasscode)
+            let ra = ReauthCoordinator(window: window,
+                                       analyticsService: analyticsService,
+                                       userStore: userStore)
+            openChildModally(ra, animated: true)
+            reauthCoordinator = ra
+        }
```
DylanRJ, line 66: Passing the action here for when the user selects to interact with the hello-world API but they're not authenticated feels weird and potentially wrong as it can't be tested effectively but seemed to be the only feasible way to do it to have it work within the coordinator pattern.

jdubey: It might make more sense to inject the HomeCoordinator to the DeveloperMenuViewController.  It's not an atypical way to use coordinators.  So something like

`weak var parentCoordinator: HomeCoordinator`

```
    private func helloWorldUnauthorizedPath() {
        if userStore.validAuthenticatedUser {
            // do stuff
        } else {
            parentCoordinator.stuffYouCurrentlyHaveAsViewModelAction()
        }
```

jdubey: Another thought:  Since in this particular case, we are ultiimately dismissing the DeveloperMenuViewController if userStore.validAuthenticatedUser is false, you could just call finish in that instance, and then handle all the clean up in performChildCleanup, etc.

Sources/Utilities/Storage/UserStorable.swift
```diff
@@ -21,22 +21,17 @@ extension UserStorable {
     }
     
     var validAuthenticatedUser: Bool {
-        guard let previouslyAuthenticatedUser else {
-            return false
-        }
+        guard let previouslyAuthenticatedUser else { return false }
```
jdubey, line 24: I think just 
`return previouslyAuthenticatedUser?.timeIntervalSinceNow.sign == .plus` 
will do the right thing and you don't need the guard statement

Sources/Application/MainCoordinator.swift
jdubey: I'm ok with using a flag, but I think it needs a better name.  requiresReauth or isReauthing or something like that.

Sources/Screens/DeveloperMenu/DeveloperMenuViewController.swift
```diff
@@ -40,18 +43,22 @@ final class DeveloperMenuViewController: BaseViewController {
...
-            do {
-                let data = try await networkClient?.makeAuthorizedRequest(exchangeRequest: URLRequest(url: AppEnvironment.stsToken),
-                                                                          scope: "sts-test.hello-world.read",
-                                                                          request: URLRequest(url: AppEnvironment.stsHelloWorld))
-                happyPathResultLabel.showSuccessMessage("Success: \(String(decoding: data!, as: UTF8.self))")
-            } catch let error as ServerError {
-                happyPathResultLabel.showErrorMessage(error)
-            } catch {
-                happyPathResultLabel.showErrorMessage()
+        if userStore.validAuthenticatedUser {
```
jdubey, line 46: If the user has opted out of biometrics, this will always fail.  I think we need to be looking at TokenHolder.shared.validAccessToken as well.
I say as well because I just tested using it *instead*, and then it fail if you opt in to biometrics, quit the app, and come back.  As the TokenHolder is a singleton now, I think what should maybe happen is that after tokens are read from the secure store we stash them in the TokenHolder and use the shared instance to reference them in all cases.
Then, if the TokenHolder becomes our source of truth, perhaps it would make sense to store requiresReauth flag there, rather than having it on MainCoordinator.

DylanRJ: Yep, good shout, forgot about non-local auth users here.
I agree TokenHolder should be our single source of truth for all the tokens.
The problem there is the fact we get our full `TokenResponse` object including all tokens, expiry date etc. from the web flow (and populate the `tokenResponse` property on TokenHolder which `validAccessToken` references).
Though, for a returning user, we only access one token at a time and store them on the top-level properties of TokenHolder, meaning the `tokenResponse` property is nil and `validAccessToken` would be false.
We've talked about this before and how we could/should be retrieving all the tokens from secure store for a returning user but maybe we need to bump the priority of the work to do that in light of this.
As a work around in lieu of that work being done I guess we could put some sort of logic for `userStore.validAuthenticatedUser` OR `TokenHolder.shared.validAccessToken`..?

jdubey: Yes, I think do an OR for now.

jdubey (approved):

mcgarj (approved):

PR #182, opened 2024-08-23
refactor: encapsulate session management into a single object

Oliver-Binns:
# Refactor of Session Management

A refactor of session management to remove duplication and improve naming between `UserStorable` and `TokenHolder` classes.

There is a new `SessionManager` type that is now responsible for parsing, storage and usage of login tokens. This class encapsulates the functionality meaning business logic and details have been removed from coordinators throughout the app.

Still todo:
- [x] Fix test compilation and build
- [x] Add additional new tests for the session manager
- [x] Ensure that secure store key generation works as expected

# Checklist

- [ ] ~~Met all accessibility requirements?~~ - No UI changes
    - [ ] ~~Checked dynamic type sizes are applied~~
    - [ ] ~~Checked VoiceOver can navigate your new code~~
    - [ ] ~~Checked a user can navigate only using a keyboard around your new code~~


Sources/Application/MainCoordinator.swift
```diff
@@ -172,9 +169,10 @@ extension MainCoordinator {
     }
     
     private func updateToken() {
-        homeCoordinator?.updateToken()
-        walletCoordinator?.updateToken()
-        profileCoordinator?.updateToken()
+        if let user = sessionManager.user {
```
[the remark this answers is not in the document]
Oliver-Binns: I would actually prefer to remove this completely and have the home / profile coordinator subscribe to updates from the SessionManager!

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -148,13 +148,17 @@ final class PersistentSessionManager: SessionManager {
             .readItem(itemName: .accessToken)
         tokenProvider.update(accessToken: accessToken)
     }
+    
+    func stopCurrentSession() {
```
obinns-dd, line 152: interesting, why "stop" rather than "end"? 🙂

Sources/Application/MainCoordinator.swift
```diff
@@ -57,19 +57,23 @@ final class MainCoordinator: NSObject,
...
-            switch error {
-            case is JWTVerifierError,
-                SecureStoreError.unableToRetrieveFromUserDefaults,
-                SecureStoreError.cantInitialiseData,
-                SecureStoreError.cantRetrieveKey:
-                fullLogin(loginError: error)
-            default:
-                print("Token retrival error: \(error)")
+        Task {
+            await MainActor.run {
```
obinns-dd, line 61: I'm not sure this is the right approach. I think we should make the `MainCoordinator` a MainActor, rather than wrapping things in a task. Coordinators should always be main actors.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
DylanRJ: Just because the method below is "end" 🙂

Sources/Application/MainCoordinator.swift
DylanRJ: Yeah, I was hoping that there was some other way to do this, I did have a play around with marking this as `MainActor` and it didn't seem to solve the app crashing after 15~ seconds with the local auth prompt displayed on the Unlock screen. Maybe one to talk about in-person

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
DylanRJ: Thinking about this some more, this isn't right as it doesn't _always_ prompt for local auth based on `sessionExists` in the `SceneDelegate` if you call this method when sending to the background.. 🤔

DylanRJ (approved):

Magdaline-A (approved):

PR #183, opened 2024-08-30
feat: update logic to build wallet tab

Magdaline-A:
# DCMAW-9857: Update logic to build (or not) Wallet tab

Add logic for each feature flag combo to see if we show it or not

# Checklist

~- [ ] Created a `draft` pull request if it is not yet ready for review~

~- [ ] Ran the app and tested the feature on a range of device sizes~
      ~Please include iPod Touch, iPhone SE and iPhone 11 as a minimum.~

~- [ ] Met all accessibility requirements?~
    ~- [ ] Checked dynamic type sizes are applied~
    ~- [ ] Checked VoiceOver can navigate your new code~
    ~- [ ] Checked a user can navigate only using a keyboard around your new code~ 


DylanRJ:
Just a thought on all of this:
Managing the tab bar is quite complicated (showing or hiding the wallet tab based on the logic tree) so wondering if it might be an idea for us to have a separate object to handle some of this rather than having that code entirely in the coordinator.

Magdaline-A:
> Just a thought on all of this: Managing the tab bar is quite complicated (showing or hiding the wallet tab based on the logic tree) so wondering if it might be an idea for us to have a separate object to handle some of this rather than having that code entirely in the coordinator.

I thought that was the point of the Wallet Availability Service.. there's only really 3 new lines of code in the coordinator

Tests/UnitTests/Application/MainCoordinatorTests.swift
```diff
@@ -124,6 +148,8 @@ extension MainCoordinatorTests {
         XCTAssertTrue(sut.childCoordinators[3] is LoginCoordinator)
         // THEN the root's delegate is the MainCoordinator
         XCTAssertTrue(sut.root.delegate === sut)
+
+        UserDefaults.standard.removeObject(forKey: FeatureFlags.enableWalletVisibleToAll.rawValue)
```
mcgarj, line 152: As this line seems to be used in most of the tests, should it be included in the teardown?

Tests/UnitTests/Application/MainCoordinatorTests.swift
```diff
@@ -17,6 +18,7 @@ final class MainCoordinatorTests: XCTestCase {
     var mockUserStore: UserStorage!
     var mockTokenVerifier: MockTokenVerifier!
     var mockUpdateService: MockAppInformationService!
+    var walletAvailabilityService = WalletAvailabilityService()
```
mcgarj, line 21: Should a mock for WalletAvailabilityService be used in MainCoordinator? Or is that not worth setting up do we think?

DylanRJ:
> > Just a thought on all of this: Managing the tab bar is quite complicated (showing or hiding the wallet tab based on the logic tree) so wondering if it might be an idea for us to have a separate object to handle some of this rather than having that code entirely in the coordinator.
> 
> I thought that was the point of the Wallet Availability Service.. there's only really 3 new lines of code in the coordinator

Yeah perhaps it isn't totally necessary, my thoughts were more-so about how we would manage adding the wallet tab on deeplinking into the app which isn't covered in this. We would need to add it to the list of view controller owned by the tab bar controller which would add some code.

Tests/UnitTests/Application/MainCoordinatorTests.swift
Magdaline-A: I don't even think I ended up using this in the end - might just need to remove it altogether 😅 Thanks!

Tests/UnitTests/Service/WalletAvailabilityServiceTests.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+@testable import OneLogin
+import XCTest
+
+final class WalletAvailabilityServiceTests: XCTestCase {
+    private var sut: MockWalletAvailabilityService!
```
DylanRJ, line 6: We need to test the actual class rather than the mock 🙂

Tests/UnitTests/Application/MainCoordinatorTests.swift
DylanRJ: In the new PR we're actually using:
```swift
AppEnvironment.updateReleaseFlags([:])
```
in tests to reset things which is a slightly nice way than calling `UserDefaults` directly

mcgarj (approved):
🙌

Tests/UnitTests/Application/MainCoordinatorTests.swift
```diff
@@ -249,13 +273,18 @@ extension MainCoordinatorTests {
...
         XCTAssertEqual(mockAnalyticsService.additionalParameters["taxonomy_level3"] as? String, "undefined")
+        
+        UserDefaults.standard.removeObject(forKey: FeatureFlags.enableWalletVisibleToAll.rawValue)
     }
     
     @MainActor
     func test_didSelect_tabBarItem_profile() {
+        UserDefaults.standard.set(false, forKey: FeatureFlags.enableWalletVisibleToAll.rawValue)
+        UserDefaults.standard.set(false, forKey: "hasAccessedWalletBefore")
+
```
DylanRJ, line 284: Can we use `AppEnvironment.updateReleaseFlags` here as we have in other places

Tests/UnitTests/Application/MainCoordinatorTests.swift
```diff
@@ -268,6 +297,9 @@ extension MainCoordinatorTests {
         XCTAssertEqual(mockAnalyticsService.eventsParamsLogged["text"], iconEvent.text)
         XCTAssertEqual(mockAnalyticsService.additionalParameters["taxonomy_level2"] as? String, "login")
         XCTAssertEqual(mockAnalyticsService.additionalParameters["taxonomy_level3"] as? String, "undefined")
+        
+        UserDefaults.standard.removeObject(forKey: FeatureFlags.enableWalletVisibleToAll.rawValue)
+        UserDefaults.standard.removeObject(forKey: "hasAccessedWalletBefore")
```
DylanRJ, line 302: And here please

Sources/Service/WalletAvailabilityService.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import Networking
+import UIKit
+
+
+public final class WalletAvailabilityService {
```
DylanRJ, line 6: I'm wondering if this needs to be a class as it's quite lightweight and would only be needed here

Sources/Service/WalletAvailabilityService.swift
```diff
@@ -0,0 +1,24 @@
...
+    let defaults = UserDefaults.standard
```
DylanRJ, line 7: Not sure how much assigning this to a variable is giving us, might make it slightly harder to read

Sources/Service/WalletAvailabilityService.swift
```diff
@@ -0,0 +1,24 @@
...
+    
+    var walletVisibleToAll = AppEnvironment.walletVisibleToAll
+    var deeplinkAccepted = AppEnvironment.walletVisibleViaDeepLink
```
DylanRJ, line 10: similarly with these

Tests/UnitTests/Application/MainCoordinatorTests.swift
Magdaline-A: So for some reason, when using `AppEnvironment.updateReleaseFlags` locally, the tests do not pass consistently..? Happy to have a proper discussion about it, I could be doing something incorrectly elsewhere.

Sources/Service/WalletAvailabilityService.swift
Magdaline-A: That's fair.. I thought it was useful more so for the `WalletAvailabilityServiceTests`. But will remove for now & try to find a better alternative for the tests 😅

Sources/Service/WalletAvailabilityService.swift
Magdaline-A: That's fair.. I thought it was useful more so for the WalletAvailabilityServiceTests. But will remove for now & try to find a better alternative for the tests 😅

DylanRJ (approved):
Well done!

OneLogin.xcworkspace/xcshareddata/swiftpm/Package.resolved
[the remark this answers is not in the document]
Magdaline-A: Yeah, I'm not sure. I had to resolve my packages at one point & I think this was the result of it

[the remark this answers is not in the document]
DylanRJ: Yeah I imagine it came from Oli's branch [here](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/185/files#diff-73671de190cf3c70aa57a72873a2d0126ab69a743516c7bfe2328effd62bed06) 🙂

PR #187, opened 2024-09-05
refactor: extract configuration from xcodeproj

obinns-dd:
# refactor: extract configuration from xcodeproj

This pull request makes no functional change.

It moves configuration outside of the `xcodeproj` file into separate `xcconfig` files.

`xcconfig` files are a better approach as they are smaller and do not often change.
This means changes that occur are easier to spot during code-reviews.

### DIFF
--- a/Sources/Application/Environment/Configuration/Debug.xcconfig
+++ b/Sources/Application/Environment/Configuration/Debug.xcconfig
@@ -1,0 +1,16 @@
+INFOPLIST_KEY_CFBundleDisplayName = One Login - Debug
+INFOPLIST_KEY_NSCameraUsageDescription = This app does not use the camera
+
+BASE_URL = mobile.staging.account.gov.uk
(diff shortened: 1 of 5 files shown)


Sources/Application/Environment/Configuration/Debug.xcconfig
```diff
@@ -0,0 +1,16 @@
+INFOPLIST_KEY_CFBundleDisplayName = One Login - Debug
+INFOPLIST_KEY_NSCameraUsageDescription = This app does not use the camera
```
obinns-dd, line 2: We need to pick this one up with content designers: we do now have some camera functionality with QR scanning 🙂

obinns-dd: Defect raised to address this: DCMAW-10156

dylanjames-dd (approved):

benmckillop (approved):

PR #191, opened 2024-09-09
feat: app qualifying

dylanjames-dd:
# DCMAW-7425: iOS | User sees 'You need to update...' page when app doesn't meet minimum app version
# DCMAW-9866: iOS | Add logic to show the screen where we have those TODO's and then test cases for each one

This PR is intended, among other things, as a refactor of the app flow from app boot and bringing the app to the foreground from the background.
Code comments are included in an attempt to describe the aims of the changes.

# Checklist


Sources/Application/QualifyingCoordinator.swift
dylanjames-dd, line 1: This coordinator along with the `AppQualifyingService` are intended to conduct app and user qualifying checks when the app is first booted and when the app is brought to the foreground from the background.

The flow of these objects and the Login and Main coordinators which constitute the rest of the app functionality is demonstrated in this simple diagram: 
image 2024-09-09 at 20 42 28

Sources/Tabs/MainCoordinator.swift
```diff
@@ -9,141 +9,47 @@ import UIKit
...
-    
-    @objc private func startReauth() {
-        fullLogin(loginError: TokenError.expired)
-    }
-    
-    private func fullLogin(loginError: Error? = nil) {
-        sessionManager.endCurrentSession()
-        showLogin(loginError)
-        windowManager.hideUnlockWindow()
+        addTabs()
```
dylanjames-dd, line 41: All calls to local auth have been removed from the MainCoordinator as this coordinator will be solely responsible for managing the tab bar functionality of the app.

Sources/Tabs/MainCoordinator.swift
```diff
@@ -9,141 +9,47 @@ import UIKit
...
-        let lc = LoginCoordinator(appWindow: windowManager.appWindow,
-                                  root: UINavigationController(),
-                                  analyticsCenter: analyticsCenter,
-                                  sessionManager: sessionManager,
-                                  networkMonitor: NetworkMonitor.shared,
-                                  loginError: loginError)
-        openChildModally(lc, animated: false)
-        loginCoordinator = lc
-    }
-
```
dylanjames-dd, line 146: There is no reference to login from the MainCoordinator class anymore as the login flow sits as a child flow to the QualifyingCoordinator. The QualifyingCoordinator should subscribe to events which may emanate from the MainCoordinator or child coordinators and present the login flow as an independent journey which depends upon state which sits across the whole app.

Sources/Utilities/AppQualifyingService.swift
```diff
@@ -0,0 +1,126 @@
...
+            }
+            delegate?.didChangeAppInfoState(state: appInfoState)
+        }
+    }
+    
+    private var userState: AppLocalAuthState = .userUnconfirmed {
+        didSet {
+            delegate?.didChangeUserState(state: userState)
+        }
+    }
```
dylanjames-dd, line 59: Using the delegate pattern these properties provide updates to their delegate class (the QualifyingCoordinator in this case) to guide the flow of app and user qualification. The app state process guides some navigation paths or, in a happy path scenario, falls through to the user state process which has it's own navigation paths.

Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -97,9 +97,17 @@ final class AppQualifyingService: QualifyingService {
                     userState = .userConfirmed
                 }
             } catch {
-                sessionManager.endCurrentSession()
-                sessionManager.clearAllSessionData()
-                userState = .userFailed(error)
+                switch error {
```
obinns-dd, line 100: is there a reason we don't want to do this for all errors?

dylanjames-dd: A `SecureStoreError.cantDecryptData` is thrown when the local auth prompt is cancelled/dismissed. I imagine from [here](https://github.com/govuk-one-login/mobile-ios-secure-store/blob/main/Sources/SecureStore/KeyManagerService.swift#L189C19-L189C51)?

obinns-dd: Gotcha, and we don't want to clear the session in that instance? I think a safer approach would be to catch that specific error and go through a different path, rather than leaving an empty generic catch block.

dylanjames-dd: My understanding was that we don't want to do anything other than leaving the user on the lock screen. When the user returns to the app either from boot or background the checks are performed again.
Happy to specifically catch that error but not sure what the body would be other than `return`?

obinns-dd: Surely an error screen and allowing them to re-login would be better than leaving them on the Lock Screen? 🤔

dylanjames-dd: An error screen when they dismiss the local authentication prompt?

obinns-dd: Ah sorry, I meant.. we catch _that_ error and return (doing nothing) and handle all _other_ errors.
Rather than what we have here now, which is catch _some_ specific errors we know about and do nothing for _other_ errors including _SecureStoreError.cantDecryptData_.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -173,12 +196,22 @@ final class PersistentSessionManager: SessionManager {
...
 
-    func clearAllSessionData() {
+    func clearAllSessionData() throws {
         encryptedStore.deleteItem(itemName: .persistentSessionID)
         unprotectedStore.removeObject(forKey: .returningUser)
         unprotectedStore.removeObject(forKey: .accessTokenExpiry)
+
+        try sessionBoundData.forEach {
+            try $0.delete()
+        }
```
dylanjames-dd, line 209: Should this be at the top of the function so if wallet data fails to delete we throw the error before deleting the above session data?

Sources/Login/LoginCoordinator.swift
```diff
@@ -11,130 +11,101 @@ final class LoginCoordinator: NSObject,
...
-                .createUnableToLoginError(errorDescription: loginError.localizedDescription,
-                                          analyticsService: analyticsCenter.analyticsService) { [unowned self] in
-                    root.dismiss(animated: true)
+
+    func authenticate() {
+//        guard sessionManager.isReauthSupported else {
+//            // DELETE DATA
+//            NotificationCenter.default.post(name: .clearWallet, object: nil)
+//            return
+//        }
```
dylanjames-dd, line 71: Guess we can get rid of this once we have the guard condition at the start of `startSession` confirmed

obinns-dd: Yes, I've removed now 🙂

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
obinns-dd: Yes, makes sense to me 🙂

Sources/Application/SceneDelegate.swift
```diff
@@ -6,68 +6,75 @@ import Logging
...
-        }
+        rootCoordinator?.lock()
     }
     
     func sceneWillEnterForeground(_ scene: UIScene) {
-        if shouldCallSceneWillEnterForeground {
-            coordinator?.evaluateRevisit()
+        guard appBooted else {
+            appBooted = true
+            return
```
Magdaline-A, line 75: Why do we need the guard if we're setting`appBooted` to true regardless?

Sources/Qualification/QualifyingCoordinator.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    func lock() {
+        displayViewController(unlockViewController)
+    }
+
+    func didChangeAppInfoState(state appInfoState: AppInformationState) {
+        switch appInfoState {
+        case .appUnconfirmed:
+            lock()
+        case .appConfirmed:
```
Magdaline-A, line 69: what do we mean by app confirmed/ unconfirmed? As in what state is the app in here?

Sources/Tabs/MainCoordinator.swift
Magdaline-A: Can we maybe get this documented somewhere that this is what the main coordinator is for now? I feel like I'd forget this 😅

Sources/Application/SceneDelegate.swift
dylanjames-dd: When the app is first booted this `sceneWillEnterForeground` is called but we don't want to call initiate on the qualifying service as this happens when the qualifying service is initialised.
Thinking about it now off the back of this question though, I don't think there's any reason _why_ we can't call it every time rather than doing it on initialisation 🙂
Will have a quick play around to see if there's any limitation to that but good shout!

Sources/Qualification/QualifyingCoordinator.swift
dylanjames-dd: App unconfirmed means we don't know what the state of the app is therefore the qualifying checks need to run and the lock screen should be displayed, this is the initial state of the `appInfoState` property in the `AppQualifyingService` before any checks have run.
App confirmed is in the inverse where we're happy that the app qualifying checks have completed successfully and we want to then move onto qualifying the user.
Does that make sense? Happy to have a chat about it or rename some stuff if that isn't immediately obvious 🙂

Sources/Tabs/MainCoordinator.swift
dylanjames-dd: Yep absolutely, we've talked about renaming this coordinator as it actually just manages the tabs now. Unfortunately "TabCoordinator" is the name of the protocol so it's been left as is but, yes, let's rename and leave a bit of DocC at the top of the class to make this obvious 🙂

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -1,11 +1,30 @@
...
+    case sessionMismatch
+    case cannotDeleteData(Error)
+    
+    static func == (lhs: PersistentSessionError, rhs: PersistentSessionError) -> Bool {
+        switch (lhs, rhs) {
+        case (.noSessionExists, .noSessionExists),
+            (.userRemovedLocalAuth, .userRemovedLocalAuth),
+            (.sessionMismatch, .sessionMismatch),
+            (.cannotDeleteData, .cannotDeleteData):
+            true
```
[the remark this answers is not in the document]
dylanjames-dd: Yep, stands for "left hand side" and "right hand side" 🙂 
This is for making the enum conform to `Equatable` so we can perform our double equals operator on it `==`. The problem here is that we have an associated value to the case `.cannotDeleteData(Error)` which means we have to implement `Equatable` ourselves (a more in depth explanation of why can probably be found somewhere).

Sources/Qualification/QualifyingCoordinator.swift
Magdaline-A: Yes, makes sense - thank you!

Magdaline-A (approved):

Sources/Qualification/QualifyingCoordinator.swift
obinns-dd: Should we rename these to be clearer?

co-eoin:
QA Approved ✅

PR #195, opened 2024-09-17
ci: update pipelines to use Xcode 16

obinns-dd:
# ci: update pipelines to use Xcode 16

Upgraded to use the latest version of Xcode.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -16,7 +16,7 @@ require 'securerandom'
 ENV["FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT"] = "600"
 
 default_platform(:ios)
-xcode_select("/Applications/Xcode_15.4.app")
+xcode_select("/Applications/Xcode_16_Release_Candidate.app")
 
 platform :ios do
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -16,7 +16,7 @@ require 'securerandom'
 ENV["FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT"] = "600"
 
 default_platform(:ios)
-xcode_select("/Applications/Xcode_15.4.app")
+xcode_select("/Applications/Xcode_16_Release_Candidate.app")
```
obinns-dd, line 19: Looks like Apple has pulled the RC and re-released a new build meaning the RC can no longer upload to App Store Connect. We now need to wait until GitHub Actions has added support for the final version of Xcode 16 before upgrading: https://github.com/actions/runner-images/issues/10629

obinns-dd: I've updated it to Xcode 16 🙂

dylanjames-dd (approved):

gabby-san-digco (approved):

PR #196, opened 2024-09-17
ci: swiftlint failures

dylanjames-dd:
# DCMAW-10233: iOS | MP | SwiftLint failing on main

Updated code to pass linting since the latest release of SwiftLint:
https://github.com/realm/SwiftLint/releases/tag/0.57.0


obinns-dd:
@dylanjames-dd: I think it would be a good idea to raise a separate ticket to pin the version of SwiftLint so this doesn't happen again: possibly by adding it as a Swift Package plugin instead?

obinns-dd (approved):

mcgarj (approved):

PR #197, opened 2024-09-17
feat: implements GA4 on SignOutWarning screen

mcgarj:
# DCMAW-9634: Implement GA4 schema on the 'You have been signed out' page

This implements the GA4 schema for the SignOutWarning screen. 

_Note_: the proposed solution in this PR has meant the `SignOutWarningViewModel` has changed from a struct to a class due to having to rework how the taxonomy is set, as the this schema requires an update to set it for 'reauth'. 

**Log for analytics:**
_Screen:_
```
11.1.0 - [FirebaseAnalytics][I-ACS023072] Event logged. Event name, event params: navigation, {
    external = false;
    ga_debug (_dbg) = 1;
    ga_event_origin (_o) = app;
    ga_realtime (_r) = 1;
    ga_screen (_sn) = you’ve been signed out;
    ga_screen_class (_sc) = signOutWarningScreen;
    ga_screen_id (_si) = 1423121875559033750;
    language = en;
    link_domain = mobile.build.account.gov.uk;
    organisation = <OT1056>;
    primary_publishing_organisation = government digital service - digital identity;
    saved_doc_type = undefined;
    taxonomy_level1 = one login mobile application;
    taxonomy_level2 = login;
    taxonomy_level3 = re auth;
    text = sign in with gov.uk one login;
    type = generic link;
}
```

_ButtonEvent:_
```
11.1.0 - [FirebaseAnalytics][I-ACS023105] Event is not subject to real-time event count daily limit. Marking an event as real-time. Event name, parameters: navigation, {
    external = false;
    ga_debug (_dbg) = 1;
    ga_event_origin (_o) = app;
    ga_realtime (_r) = 1;
    ga_screen (_sn) = you’ve been signed out;
    ga_screen_class (_sc) = signOutWarningScreen;
    ga_screen_id (_si) = 1423121875559033750;
    language = en;
    link_domain = mobile.build.account.gov.uk;
    organisation = <OT1056>;
    primary_publishing_organisation = government digital service - digital identity;
    saved_doc_type = undefined;
    taxonomy_level1 = one login mobile application;
    taxonomy_level2 = login;
    taxonomy_level3 = re auth;
    text = sign in with gov.uk one login;
    type = generic link;
}
```
The screenID is given the value here: `case signOutWarning = "cfc50baa-4b56-4170-9707-cd05b60b6658"` in the `ErrorAnalyticsScreenID` enum in the `ErrorAnalytics` file

**Analytics disabled:**
```
11.1.0 - [FirebaseAnalytics][I-ACS023049] Analytics is disabled. Event not logged
11.1.0 - [FirebaseAnalytics][I-ACS023049] Analytics is disabled. Event not logged
```

**Welsh:**
_Screen:_
```
11.1.0 - [FirebaseAnalytics][I-ACS023105] Event is not subject to real-time event count daily limit. Marking an event as real-time. Event name, parameters: screen_view (_vs), {
    _mst = 1;
    ga_debug (_dbg) = 1;
    ga_event_origin (_o) = app;
    ga_previous_class (_pc) = introWelcomeScreen;
    ga_previous_id (_pi) = -3966439220278464730;
    ga_previous_screen (_pn) = gov.uk one login;
    ga_realtime (_r) = 1;
    ga_screen (_sn) = you’ve been signed out;
    ga_screen_class (_sc) = signOutWarningScreen;
    ga_screen_id (_si) = -3966439220278464729;
    language = cy;
    organisation = <OT1056>;
    primary_publishing_organisation = government digital service - digital identity;
    saved_doc_type = undefined;
    screen_id = signoutwarningscreen;
    taxonomy_level1 = one login mobile application;
    taxonomy_level2 = login;
    taxonomy_level3 = re auth;
    title = you’ve been signed out;
}
```

_Button:_
```
11.1.0 - [FirebaseAnalytics][I-ACS023105] Event is not subject to real-time event count daily limit. Marking an event as real-time. Event name, parameters: navigation, {
    external = false;
    ga_debug (_dbg) = 1;
    ga_event_origin (_o) = app;
    ga_realtime (_r) = 1;
    ga_screen (_sn) = you’ve been signed out;
    ga_screen_class (_sc) = signOutWarningScreen;
    ga_screen_id (_si) = -3966439220278464729;
    language = cy;
    link_domain = mobile.build.account.gov.uk;
    organisation = <OT1056>;
    primary_publishing_organisation = government digital service - digital identity;
    saved_doc_type = undefined;
    taxonomy_level1 = one login mobile application;
    taxonomy_level2 = login;
    taxonomy_level3 = re auth;
    text = sign in with gov.uk one login;
    type = generic link;
}
```

# Checklist

### DIFF
--- a/Sources/Errors/SignOutWarningViewModel.swift
+++ b/Sources/Errors/SignOutWarningViewModel.swift
@@ -2,19 +2,21 @@ import GDSAnalytics
 import GDSCommon
 import Logging
 
-class SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
+struct SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
     let title: GDSLocalisedString = "app_signOutWarningTitle"
     let body: GDSLocalisedString = "app_signOutWarningBody"
     let primaryButtonViewModel: ButtonViewModel
     let secondaryButtonViewModel: ButtonViewModel? = nil
-    var analyticsService: AnalyticsService
+    let analyticsService: AnalyticsService
 
     let rightBarButtonTitle: GDSLocalisedString? = nil
     let backButtonIsHidden: Bool = true
     
     init(analyticsService: AnalyticsService,
          action: @escaping () -> Void) {
-        self.analyticsService = analyticsService
+        var tempAnalyticsService = analyticsService
+        tempAnalyticsService.setAdditionalParameters(appTaxonomy: .reauth)
+        self.analyticsService = tempAnalyticsService
         let event = LinkEvent(textKey: "app_extendedSignInButton",
                               linkDomain: AppEnvironment.oneLoginBaseURL,
                               external: .false)
@@ -26,7 +28,6 @@ class SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
     }
     
     func didAppear() {
-        analyticsService.setAdditionalParameters(appTaxonomy: .reauth)
         let screen = ScreenView(id: ErrorAnalyticsScreen.signOutWarning.rawValue,
                                 screen: ErrorAnalyticsScreen.signOutWarning,
                                 titleKey: title.stringKey)
--- a/Sources/Errors/SignOutWarningViewModel.swift
+++ b/Sources/Errors/SignOutWarningViewModel.swift
@@ -29,7 +29,7 @@ class SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
         analyticsService.setAdditionalParameters(appTaxonomy: .reauth)
         let screen = ScreenView(id: ErrorAnalyticsScreen.signOutWarning.rawValue,
                                 screen: ErrorAnalyticsScreen.signOutWarning,
-                                titleKey: "app_signOutWarningTitle")
+                                titleKey: title.stringKey)
         analyticsService.trackScreen(screen)
     }
     
--- a/Sources/Errors/SignOutWarningViewModel.swift
+++ b/Sources/Errors/SignOutWarningViewModel.swift
@@ -15,13 +15,22 @@ struct SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
     init(analyticsService: AnalyticsService,
          action: @escaping () -> Void) {
         self.analyticsService = analyticsService
+        let event = LinkEvent(textKey: "app_extendedSignInButton",
+                              linkDomain: AppEnvironment.oneLoginBaseURL,
+                              external: .false)
         self.primaryButtonViewModel = AnalyticsButtonViewModel(titleKey: "app_extendedSignInButton",
-                                                               analyticsService: analyticsService) {
+                                                               analyticsService: analyticsService,
+                                                               analyticsEvent: event) {
             action()
         }
     }
     
-    func didAppear() { /* BaseViewModel compliance */ }
+    func didAppear() {
+        let screen = ScreenView(id: ErrorAnalyticsScreen.signOutWarning.rawValue,
+                                screen: ErrorAnalyticsScreen.signOutWarning,
+                                titleKey: "app_signOutWarningTitle")
+        analyticsService.trackScreen(screen)
+    }
     
     func didDismiss() { /* BaseViewModel compliance */ }
 }
--- a/Sources/Errors/SignOutWarningViewModel.swift
+++ b/Sources/Errors/SignOutWarningViewModel.swift
@@ -2,13 +2,13 @@ import GDSAnalytics
 import GDSCommon
 import Logging
 
-struct SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
+class SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
     let title: GDSLocalisedString = "app_signOutWarningTitle"
     let body: GDSLocalisedString = "app_signOutWarningBody"
     let primaryButtonViewModel: ButtonViewModel
     let secondaryButtonViewModel: ButtonViewModel? = nil
-    let analyticsService: AnalyticsService
-    
+    var analyticsService: AnalyticsService
+
     let rightBarButtonTitle: GDSLocalisedString? = nil
     let backButtonIsHidden: Bool = true
     
@@ -26,6 +26,7 @@ struct SignOutWarningViewModel: GDSErrorViewModelV2, BaseViewModel {
     }
     
     func didAppear() {
+        analyticsService.setAdditionalParameters(appTaxonomy: .reauth)
         let screen = ScreenView(id: ErrorAnalyticsScreen.signOutWarning.rawValue,
                                 screen: ErrorAnalyticsScreen.signOutWarning,
                                 titleKey: "app_signOutWarningTitle")
--- a/Sources/Extensions/CustomTypes/AnalyticsService+GDS.swift
+++ b/Sources/Extensions/CustomTypes/AnalyticsService+GDS.swift
@@ -8,6 +8,7 @@ enum AppTaxonomy: String {
     case home
     case wallet
     case profile
+    case reauth = "re auth"
 }
 
 extension AnalyticsService {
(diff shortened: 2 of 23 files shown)


Sources/Errors/SignOutWarningViewModel.swift
```diff
@@ -2,26 +2,36 @@ import GDSAnalytics
...
             action()
         }
     }
     
-    func didAppear() { /* BaseViewModel compliance */ }
+    func didAppear() {
+        analyticsService.setAdditionalParameters(appTaxonomy: .reauth)
+        let screen = ScreenView(id: ErrorAnalyticsScreen.signOutWarning.rawValue,
+                                screen: ErrorAnalyticsScreen.signOutWarning,
+                                titleKey: "app_signOutWarningTitle")
```
benmckillop, line 32: Could we use `title.stringKey` here instead of the static string please :)

Sources/Extensions/CustomTypes/AnalyticsService+GDS.swift
```diff
@@ -8,6 +8,7 @@ enum AppTaxonomy: String {
     case home
     case wallet
     case profile
+    case reauth = "re auth"
```
benmckillop, line 11: should this have a space in it? maybe something we need to mention to the analytics team to see if we can avoid

dylanjames-dd:
I wonder if we should make a change to the Logging package to make the `AnalyticsService` protocol inherit from `NSObject` which will ensure any conforming types will have to be reference types.
I've created a [branch here](https://github.com/govuk-one-login/mobile-ios-logging/tree/feature/dcmaw-9634-ensuring-analyticsService-class) for the change, seems to silence the warnings and means we can have this as a `struct` and the property as a `let`.
Feels like something worth doing but will defer to @benmckillop & @obinns-dd

obinns-dd:
> [branch here](https://github.com/govuk-one-login/mobile-ios-logging/tree/feature/dcmaw-9634-ensuring-analyticsService-class)

@dylanjames-dd is there a reason we have this as a class?

I think it **should** be a struct, then we can change values on additional parameters that are logged while we pass it down the app hierarchy, without affecting top level objects.

Tests/UnitTests/Errors/SignOutWarningViewModelTests.swift
```diff
@@ -43,7 +44,29 @@ extension SignOutWarningViewModelTests {
...
         XCTAssertFalse(didCallButtonAction)
+        XCTAssertEqual(mockAnalyticsService.eventsLogged.count, 0)
         sut.primaryButtonViewModel.action()
         XCTAssertTrue(didCallButtonAction)
+        XCTAssertEqual(mockAnalyticsService.eventsLogged.count, 1)
+        let event = LinkEvent(textKey: "app_extendedSignInButton",
+                              linkDomain: AppEnvironment.oneLoginBaseURL,
+                              external: .false)
+        XCTAssertEqual(mockAnalyticsService.eventsLogged, [event.name.name])
+        XCTAssertEqual(mockAnalyticsService.eventsParamsLogged["text"], event.parameters["text"])
```
obinns-dd, line 55: Please can we not check each of the parameters individually? This tightly couples our tests to the implementation of each event type, which we don't want to do. For example, if we were to then change the implementation of the event in the logging package, we would then break all our unit tests.

We should just check that `eventParamsLogged` equals `event.parameters` 🙂

mcgarj: Sure can do! This copies how we've been testing `eventParamsLogged` in our other tests, do they need a tech debt ticket to correct given that it's technically outside the scope of this ticket?

obinns-dd: Yes, can we correct these ones here and then raise a ticket for the others please?

mcgarj: Will do! Should `screenParamsLogged` also be updated to reflect the above changes, or is that fine being like `XCTAssertEqual(mockAnalyticsService.screenParamsLogged["title"], screen.parameters["title"])`?

obinns-dd: Yeah, I think we should do the same if we can. The tests should be flexible such that if we rename the parameters in the login package, then the tests here don't break - the tests in the logging package already cover that.

Sources/Extensions/CustomTypes/AnalyticsService+GDS.swift
mcgarj: For this comment and the one about "my profile", I've started a thread with the analytics team on [Slack](https://gds.slack.com/archives/C068SNY9UFP/p1726650547125459)

Tests/UnitTests/Errors/SignOutWarningViewModelTests.swift
mcgarj: I've updated both `eventParamsLogged` and `screenParamsLogged`. Is this change what you had envisioned?

Tests/UnitTests/Mocks/Sessions+Services/Analytics/MockAnalyticsService.swift
```diff
@@ -37,7 +37,9 @@ final class MockAnalyticsService: AnalyticsService {
         parameters["AnalyticsParameterScreenClass"] = screen.type.name
         parameters["AnalyticsParameterScreenName"] = screen.name
         
-        screenParamsLogged = parameters
+        screenParamsLogged["title"] = parameters["title"]
```
obinns-dd, line 40: 🤔 I'm not sure this change is right, we should track all the parameters, surely? Otherwise the test should fail!

mcgarj: There's likely a more elegant approach, but before with the original thing we had, more parameters were being added, but the`ScreenView` only cares about the title and screen ID, if it's an error we also need the reason. 

This is the failing error I got at the time when I ran it pre-change if it helps explain the above/my thinking: 
`XCTAssertEqual failed: ("["AnalyticsParameterScreenName": "you've been signed out", "AnalyticsParameterScreenClass": "signOutWarningScreen", "screen_id": "signoutwarningscreen", "title": "you've been signed out"]") is not equal to ("["title": "you've been signed out", "screen_id": "signoutwarningscreen"]")`

mcgarj: Here's the ScreenParameters inside the actual `ScreenView` struct for easy reference: 
```
    public var parameters: [String: String] {
        [
            ScreenParameter.id.rawValue: id,
            ScreenParameter.title.rawValue: title
        ].compactMapValues(\.?.formattedAsParameter)
    }
```

obinns-dd: If we have more parameters being added, then that sounds like the test is failing for good reason. I'd suggest we should look into the root cause of the extra parameters 🙂

mcgarj: Maybe I misunderstood the error message, but weren't the extra params being added by the Mock (as it was adding all of the params), but the expected outcome in the real dictionary was `["title": "you've been signed out", "screen_id": "signoutwarningscreen"]`?

obinns-dd: hmm 👀 why are we adding extra parameters in the mock? That doesn't sound right 🤔 
The mock should just be for validating we're passing in the right parameters.

mcgarj: I'm not sure why we were initially, but it would explain the original testing approach. My update to the mock should now be passing in only the 3 parameters the tests were failing against. I did raise a tech debt ticket for this yesterday, to update the rest of the tests so I could expand on what the ticket is for to investigate/update how the mock handles these parameters?

dylanjames-dd:
> > [branch here](https://github.com/govuk-one-login/mobile-ios-logging/tree/feature/dcmaw-9634-ensuring-analyticsService-class)
> 
> @dylanjames-dd is there a reason we have this as a class?
> 
> I think it **should** be a struct, then we can change values on additional parameters that are logged while we pass it down the app hierarchy, without affecting top level objects.

Tech debt ticket for that change to the Logging package [here](https://govukverify.atlassian.net/browse/DCMAW-10246)

Sources/Extensions/CustomTypes/AnalyticsService+GDS.swift
```diff
@@ -23,9 +24,18 @@ extension AnalyticsService {
     }
     
     mutating func setAdditionalParameters(appTaxonomy: AppTaxonomy) {
+        var taxonomyLevel3: String {
```
obinns-dd, line 27: This is a real [open-closed principle](https://en.wikipedia.org/wiki/Open–closed_principle) violation for now, but I think we can let it through given we're going to address it with the Logging package change: https://govukverify.atlassian.net/browse/DCMAW-10246

dylanjames-dd (approved):

benmckillop (approved):

PR #230, opened 2024-11-07
feat: make content tile tap-able

Magdaline-A:
# DCMAW-10253: Make the 'Your Services' tile tap-able

Ensuring the ‘Your services’ tile can be tapped as a whole - tapping anywhere on the tile should behave in the same way as tapping the 'Go to your services' link currently on the tile.

### QA Evidence:
https://github.com/user-attachments/assets/d0ba66ea-51ae-4917-8293-d50bebcac283

# Checklist

~- [ ] Created a `draft` pull request if it is not yet ready for review~

### DIFF
--- a/Sources/Screens/Tabs/HomeViewController.swift
+++ b/Sources/Screens/Tabs/HomeViewController.swift
@@ -59,10 +59,6 @@ extension HomeViewController {
     }
     
     override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-        didTapCard()
-    }
-    
-    func didTapCard() {
         viewModel.cardTapped()
     }
 }
--- a/Sources/Screens/Tabs/HomeViewController.swift
+++ b/Sources/Screens/Tabs/HomeViewController.swift
@@ -6,11 +6,14 @@ import UIKit
 final class HomeViewController: UITableViewController {
     let analyticsService: AnalyticsService
     let navigationTitle: GDSLocalisedString = "app_homeTitle"
+    let viewModel: ServicesTileViewModel
 
-    init(analyticsService: AnalyticsService) {
+    init(analyticsService: AnalyticsService,
+         viewModel: ServicesTileViewModel) {
         var tempAnalyticsService = analyticsService
         tempAnalyticsService.setAdditionalParameters(appTaxonomy: .home)
         self.analyticsService = tempAnalyticsService
+        self.viewModel = viewModel
         super.init(style: .insetGrouped)
     }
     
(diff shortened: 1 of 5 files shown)


Sources/Screens/Tabs/HomeViewController.swift
```diff
@@ -54,6 +57,14 @@ extension HomeViewController {
         }
         return cell
     }
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        didTapCard()
```
robinphillips, line 62: Is there a reason for not doing `viewModel.cardTapped()` here directly rather than wrapping it in the `didTapCard` function and then calling that?

Sources/Tabs/HomeCoordinator.swift
```diff
@@ -24,7 +24,10 @@ final class HomeCoordinator: NSObject,
         root.tabBarItem = UITabBarItem(title: GDSLocalisedString(stringLiteral: "app_homeTitle").value,
                                        image: UIImage(systemName: "house"),
                                        tag: 0)
-        let hc = HomeViewController(analyticsService: analyticsService)
+        let viewModel = ServicesTileViewModel.yourServices(analyticsService: analyticsService,
+                                                           urlOpener: UIApplication.shared)
```
robinphillips, line 28: Is it worth considering defaulting the argument for `urlOpener` in the `yourServices` function to this? Eg:

```
func yourServices(analyticsService: AnalyticsService,
                  urlOpener: URLOpener = UIApplication.shared)
```

Sources/Screens/Tabs/HomeViewController.swift
Magdaline-A: Not really 😅 I can change it

Sources/Tabs/HomeCoordinator.swift
Magdaline-A: Possibly.. I tried this & got a `Main actor-isolated class property 'shared' can not be referenced from a nonisolated context...` warning. I'm assuming this means I'd have to add a `@MainActor` annotation wherever its used?

dylanjames-dd:
Question on this:
Does the row being tappable mean that the button - if it has a different action associated - would never be triggered by interacting with it because the row being tappable takes precedence over the button? Just something I've been wondering about 🙂

robinphillips:
> Question on this: Does the row being tappable mean that the button - if it has a different action associated - would never be triggered by interacting with it because the row being tappable takes precedence over the button? Just something I've been wondering about 🙂

I was wondering this too but only an issue if the actions are intended to be different. For example:
cell tap -> A
button tap -> B

vs 

cell tap -> A
button tap -> A

Sources/Tabs/HomeCoordinator.swift
robinphillips: Yeah, would make sense for this to be added to the protocol in common.

PR #232, opened 2024-11-18
feat: generating attestation proof of possession

Magdaline-A:
# DCMAW-10547: App Integrity | Generate Client Attestation PoP JWT

The app needs to generate a Proof of Possession (PoP) JWT to prove that it holds a valid key pair and can sign requests with its private key. This PR aims to set up the logic for generating the PoP in the AppIntegrity package, which will be used within the app. The logic to sign requests can be found in the [Networking repository](https://github.com/govuk-one-login/mobile-ios-networking/pull/31).

[Xcode 16 has been taken off the macos14 runners](https://github.com/actions/runner-images/issues/10703). This PR updates the macOS version used by the runners to MacOS 15.1 Sequoia.

# Checklist

~- [ ] Ran the app and tested the feature on a range of device sizes~
      ~Please include iPod Touch, iPhone SE and iPhone 11 as a minimum.~

~- [ ] Met all accessibility requirements?~
    ~- [ ] Checked dynamic type sizes are applied~
    ~- [ ] Checked VoiceOver can navigate your new code~
    ~- [ ] Checked a user can navigate only using a keyboard around your new code~ 


AppIntegrity/Sources/AppIntegrity/AttestationPOP/JWTHeader.swift
dylanjames-dd, line 1: Been having a little think about this and `JWTHeader` and the `JWTPayload` type, I think the thinking is generally right but it limits the data that we can pass through to the `generateJWT` method.
I think we should have a protocol for an expected type and probably construct and pass a concrete type in the consumer (app) space so that it is extendable in future.
Let's have a chat about it 🙂

AppIntegrity/Sources/AppIntegrity/JWTGenerator.swift
```diff
@@ -0,0 +1,3 @@
+public protocol JWTGenerator {
+    func generateJWT(header: [String: String], payload: [String: Any]) -> String
```
dylanjames-dd, line 2: Think we probably want `[String: Any]` as the expected type for the header parameter 🙂 as well

AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift
```diff
@@ -44,24 +48,28 @@ public final class FirebaseAppIntegrityService: AppIntegrityProvider {
...
+         jwtGenerator: JWTGenerator) {
         self.client = client
         self.vendor = vendor
         self.baseURL = baseURL
         self.proofOfPossessionProvider = proofOfPossessionProvider
+        self.jwtGenerator = jwtGenerator
     }
 
     public convenience init(
         client: NetworkClient,
```
dylanjames-dd, line 61: I feel like this property should be called `networkClient` to avoid any confusion 🙂

AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift
```diff
@@ -41,43 +42,48 @@ public final class FirebaseAppIntegrityService: AppIntegrityProvider {
...
         }
 
         let token = try await vendor.token(forcingRefresh: false)
-        print("vendor token: \(token.token)")
 
         do {
-            let attestation = try await fetchClientAttestation(appCheckToken: token.token)
             // TODO: DCMAW-10322 | store this locally
-            return attestation.attestationJWT
+            let attestation = try await fetchClientAttestation(appCheckToken: token.token)
```
benmckillop-dd, line 79: Wonder is there potentially better naming for the token so that we don't have `token.token`?

dylanjames-dd (approved):
🙌

PR #234, opened 2024-11-27
feat: app integrity flow

dylanjames-dd:
# feat: app integrity flow

This PR ties together work undertaken in packages this project is dependent on:
[Exposing a signing service](https://github.com/govuk-one-login/mobile-ios-secure-store/pull/26)
[Encoding the signing service public key into a JWK](https://github.com/govuk-one-login/mobile-ios-secure-store/pull/29)
[Creating a JWT](https://github.com/govuk-one-login/mobile-ios-networking/pull/31)
[Passing headers to /token endpoint](https://github.com/govuk-one-login/mobile-ios-authentication/pull/58)
In order to facilitate the App Integrity package to optionally perform an integrity check of the app when conducting a web authentication journey.

This PR also includes the work from [DCMAW-7461](https://govukverify.atlassian.net/browse/DCMAW-7461) to remove the feature flag to call auth directly rather than STS as their authentication service does not support App Attestation.

# Checklist

[DCMAW-7461]: https://govukverify.atlassian.net/browse/DCMAW-7461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Magdaline-A (approved):

.github/workflows/merged.yml
```diff
@@ -179,7 +179,6 @@ jobs:
           bundle install
 
           bundle exec fastlane testWithoutCoverageForUITests scheme:"OneLoginBuild" \
-            configuration:"Build" \
```
obinns-dd, line 182: Assume this is needed to run the tests in the Debug mode?

Sources/Application/Environment/AppEnvironment.swift
```diff
@@ -48,116 +77,146 @@ public final class AppEnvironment {
...
     }
-}
-
-class ReleaseFlags: FeatureFlagProvider {
-    var flags: [String: Bool] = [:]
-
-    subscript(key: String) -> Any? {
-        flags[key]
+    
+    class FeatureFlags: FeatureFlagProvider {
```
obinns-dd, line 88: I think we should probably think about separating these out into another file.

Sources/Application/Environment/Configuration/Debug.xcconfig
```diff
@@ -5,17 +5,12 @@ PRODUCT_BUNDLE_IDENTIFIER = uk.gov.onelogin.build
...
-FEATURE_FLAG_FILE = FeatureFlagsBuild
-
-ONE_LOGIN_AUTHORIZE_URL = auth-stub.mobile.build.account.gov.uk
-ONE_LOGIN_CLIENT_ID = TEST_CLIENT_ID
-
-REDIRECT_URL = https:/$()/mobile.build.account.gov.uk/redirect
-STS_BASE_URL = token.build.account.gov.uk
 STS_CLIENT_ID = bYrcuRVvnylvEgYSSbBjwXzHrwJ
 
-WALLET_CREDENTIAL_ISSUER_URL = https:/$()/example-credential-issuer.mobile.build.account.gov.uk
```
obinns-dd, line 20: Do we not need the issuer URL anymore?

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -125,17 +129,24 @@ final class PersistentSessionManager: SessionManager {
             throw PersistentSessionError.sessionMismatch
         }
         
-        let configuration = LoginSessionConfiguration
-            .oneLogin(persistentSessionId: persistentID)
+        let sessionConfiguration: LoginSessionConfiguration = if AppEnvironment.appIntegrityEnabled {
```
obinns-dd, line 132: should we inject the configuration in so that the `PersistentSessionManager` supports any login configuration?

Sources/Extensions/CustomTypes/LoginSessionConfiguration+OneLogin.swift
```diff
@@ -1,15 +1,42 @@
+import AppIntegrity
 import Authentication
 import Foundation
 
+protocol LoginSessionConfigurationProvider {
```
obinns-dd, line 5: I'm not sure what value this type gives, I think we should just pass in the configuration into the `PersistentSessionManager`.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
dylanjames-dd: The `AppAuthSession` only supports a `LoginSessionConfiguration` though?

Sources/Utilities/Storage/Session/SessionManager.swift
```diff
@@ -21,7 +21,7 @@ protocol SessionManager: UserProvider {
     var localAuthentication: LocalAuthenticationManager { get }
 
     /// Starts a new session from a remote login
-    func startSession(using session: LoginSession) async throws
+    func startSession(using session: LoginSession, configurationInitialiser: LoginSessionConfigurationProvider.Type) async throws
```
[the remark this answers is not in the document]
dylanjames-dd: We need to pass the `persistentID` into the `LoginSessionConfiguration` though, I suppose we could make it as part of the protocol `LoginSession` and create the `LoginSessionConfiguration` in the `AuthenticationCoordinator` but yeah.
This felt cleaner because we can pass a mock to create whatever configuration we want

.github/workflows/merged.yml
dylanjames-dd: Yep 🙂

Sources/Extensions/CustomTypes/LoginSessionConfiguration+OneLogin.swift
dylanjames-dd: See comment above regarding `persistentID` which comes from the session manager.

Sources/Application/Environment/AppEnvironment.swift
dylanjames-dd: I guess the good thing about these being within the body of `AppEnvironment` is that they can't be referenced from outside it where they should never be needed, maybe within an extension in another file?

Sources/Application/Environment/Configuration/Debug.xcconfig
dylanjames-dd: It's passed into the `WalletSDK.start` method on the current wallet version we're using 🤷‍♂️

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
obinns-dd: Yes, but we can have any instance of that type.

Sources/Utilities/Storage/Session/SessionManager.swift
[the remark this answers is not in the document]
obinns-dd: Hm 🤔 I do think it makes more sense to inject in from the coordinator because we have the flexibility without needing the complexity of different types. It feels a bit overly complex to need a protocol just to create an instance of a configuration, but not a hill I am willing to die on 🙂 whatever makes sense to you & Magdaline!

Sources/Application/Environment/Configuration/Debug.xcconfig
obinns-dd: So it's still needed?

MobilePlatformServices/Sources/MobilePlatformServices/App Info/Response Objects/App.swift
```diff
@@ -5,25 +5,31 @@ public struct App: Decodable {
     public let minimumVersion: Version
     public let allowAppUsage: Bool
     public let releaseFlags: [String: Bool]
+    public let featureFlags: [String: Bool]
```
obinns-dd, line 8: A thought for the future, it feels like some of this PR (for example: this addition of feature flags) could / should have been completed in another pull request and merged in separately 🙂 just to keep our changes smaller. That makes it easier to revert if we break anything.

Sources/Utilities/Storage/Session/SessionManager.swift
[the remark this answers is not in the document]
dylanjames-dd: I totally agree, it would be way nicer if we could just pass the instance of `LoginSessionConfiguration` to this `PersistentSessionManager.startSession()` method.
Though, an instance of `LoginSessionConfiguration` needs to be initialised with a `persistentSessionID`, that property only exists on the `PersistentSessionManager` (the class we're passing this instance of `LoginSessionConfiguration` into) so unless we go and make that property a variable in the authentication package and set it in on the instance within the body of `startSession()` then I don't know how we would be able to get the `persistentSessionID` into the `LoginSessionConfiguration` used for the login session?
The way it's currently set up where we pass in a type which can initialise a configuration felt good because we can mock the methods to provide a mock configuration instance. Totally happy to change it if we think there's a better way but I don't think we can pass in an instance of `LoginSessionConfiguration` in it's current form?

Sources/Application/Environment/Configuration/Debug.xcconfig
dylanjames-dd: For our current version of wallet (5.0.0) it is I think, yes 🙂

MobilePlatformServices/Sources/MobilePlatformServices/App Info/Response Objects/App.swift
dylanjames-dd: Agreed, I think the ticket for the feature flag change as a specific piece of work got closed.
Happy to break this stuff into a separate PR and get that merged before this. I started pulling on the AppEnvironment stuff as a piece of string and led to a lot of cleaning up in there and the configuration files, as I say, happy to take that stuff out and put in a separate PR if we think this is too much for one PR 👍

obinns-dd: I think leave it for now as it's probably more complex to split it up, but one for us to be careful of next time 🙂

Sources/Application/Environment/AppEnvironment.swift
```diff
@@ -47,117 +76,122 @@ public final class AppEnvironment {
...
         return components.url!
     }
-
-    static var privacyPolicyURL: URL {
+    
+    static var walletCredentialIssuer: URL {
         var components = URLComponents()
         components.scheme = "https"
-        components.host = string(for: .externalBaseURL)
-        components.query = "lng=\(isLocaleWelsh ? "cy" : "en")"
```
obinns-dd, line 87: does this mean the credential issuer doesn't add the welsh parameter anymore?

obinns-dd (approved):

Sources/Application/Environment/AppEnvironment.swift
dylanjames-dd: Sorry this file has be rearranged a bit now, you're looking at the privacy policy URL here which remains but is further down in this file 🙂 
The credential issuer never had the welsh parameter and still doesn't 🙂

PR #265, opened 2025-02-17
feat: settings page

Magdaline-A:
# DCMAW-10442: Settings page redesign

Updating the current 'Profile page' to the new Settings redesign.

# Checklist


Sources/Application/Environment/AppEnvironment.swift
```diff
@@ -174,6 +174,49 @@ extension AppEnvironment {
...
+        return components.url!
+    }
+    
+    static var manageAccountURLWelsh: URL {
+        var components = URLComponents()
+        components.scheme = "https"
+        components.host = govURLString
+        components.path = "/defnyddio-eich-gov-uk-one-login"
+        return components.url!
+    }
```
dylanjames-dd, line 200: Think these could probably be combined as such which will make the call sight simpler 🙂 
```suggestion
    static var manageAccountURL: URL {
        var components = URLComponents()
        components.scheme = "https"
        components.host = govURLString
        components.path = isLocaleWelsh ? "/defnyddio-eich-gov-uk-one-login" : "/using-your-gov-uk-one-login"
        return components.url!
    }
```

Sources/Tabs/SettingsCoordinator.swift
```diff
@@ -88,4 +92,12 @@ final class ProfileCoordinator: NSObject,
         let navController = UINavigationController(rootViewController: devMenuViewController)
         root.present(navController, animated: true)
     }
+    
+    func openManageDetailsPage() {
+        if AppEnvironment.isLocaleWelsh {
+            urlOpener.open(url: AppEnvironment.manageAccountURLWelsh)
+        } else {
+            urlOpener.open(url: AppEnvironment.manageAccountURL)
+        }
```
dylanjames-dd, line 101: Considering the above suggested change, you don't have to worry about the locale here and it just becomes 
```swift
urlOpener.open(url: AppEnvironment.manageAccountURL)
```
🙂

Sources/Analytics/Tabs/Profile/ProfileTabAnalytics.swift
dylanjames-dd, line 1: Think we want to rename this file name 🙂

Sources/Screens/Tabs/TabbedViewController.swift
```diff
@@ -64,20 +50,6 @@ final class TabbedViewController: BaseViewController {
         }
     }
     
-    private func subscribeToUsers() {
-        userProvider.user
-            .receive(on: DispatchQueue.main)
-            .sink { user in
-                self.updateEmail(user?.email)
-            }.store(in: &cancellables)
-    }
```
dylanjames-dd, line 73: I'm slightly worried that removing this Combine stuff will stop the updating of the user's email information. Setting to empty on sign out and resetting to a new value for the next logged in user for a new log in might be broken if we remove some of this stuff.
One I'd be interested to get the take of @obinns-dd about tbh

obinns-dd: Ah yes, without subscribing to changes the email does update in the session manager, but this update might not be copied across to the relevant view, so the view would contain the correct details when it's first created (app launch) but then not updated later if things change.

amishpatel1-gds: when the does the users email change? asking because from what i could see if a different user is logged in, then it clears the previous user and `TabManagerCoordinator` listens for a user logging out via notification and calls finish. So im assuming this view controller will be initialised again so it should always have the correct email address right?

Sources/Screens/Tabs/TabbedViewController.swift
```diff
@@ -71,11 +56,17 @@ final class TabbedViewController: BaseViewController {
                 self.updateEmail(user?.email)
             }.store(in: &cancellables)
     }
-
+    
     func updateEmail(_ email: String?) {
```
dylanjames-dd, line 60: There's no actual reason to be accepting a parameter here anymore, maybe we should just remove 🙂

amishpatel1-gds (approved):

PR #268, opened 2025-02-21
fix: refresh secure store for new user to ensure correct access

dylanjames-dd:
# fix: refresh secure store for new user to ensure correct access

This PR fixes some strange behaviour around local auth method used for accessing the secure store for a user who has not previously logged into the app.
This change refreshes the secure just before saving new login tokens if there are not any already in the instance of secure store, like for a reauth user.

# Checklist


Sources/Resources/OLString.swift
Magdaline-A, line 1: How come we've changed this to an enum now? What are the benefits of this over what we had before?

Magdaline-A (approved):

Sources/Resources/OLString.swift
dylanjames-dd: Good question 🙂 
I think I always suspected it was slightly lazy and perhaps problematic to just extend string with static let properties like we were doing before. Then I saw [this YouTube video](https://www.youtube.com/watch?v=eCMDK_XhRXg) which laid out some of the reasons to change what we're doing, I'd recommend giving it a watch (will do a better job of explaining than me 🙂

amishpatel1-gds (approved):

PR #323, opened 2025-05-01
refactor: remove deprecated GDSErrorViewController

chionia-gds:
# DCMAW-12993 Remove deprecated GDSErrorViewController

Refactored screens examples: 

# Checklist

~~- [ ] Created a `draft` pull request if it is not yet ready for review~~

### DIFF
--- a/Sources/Screens/Errors/DataDeletedWarningViewModel.swift
+++ b/Sources/Screens/Errors/DataDeletedWarningViewModel.swift
@@ -2,30 +2,29 @@ import GDSAnalytics
 import GDSCommon
 import Logging
 
-struct DataDeletedWarningViewModel: GDSErrorViewModelV2,
-                                    GDSErrorViewModelWithImage,
+struct DataDeletedWarningViewModel: GDSErrorViewModelV3,
                                     BaseViewModel {
-    let image: String = "exclamationmark.circle"
     let title: GDSLocalisedString = "app_dataDeletionWarningTitle"
-    
-    var body: GDSLocalisedString {
+    var bodyContent: [ScreenBodyItem] {
         if WalletAvailabilityService.shouldShowFeature {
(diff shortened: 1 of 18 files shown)


Sources/Screens/Errors/DataDeletedWarningViewModel.swift
```diff
@@ -2,30 +2,29 @@ import GDSAnalytics
...
     
     let rightBarButtonTitle: GDSLocalisedString? = nil
     let backButtonIsHidden: Bool = true
     
     init(action: @escaping () -> Void) {
-        self.primaryButtonViewModel = StandardButtonViewModel(titleKey: "app_signInButton") {
-            action()
-        }
+        self.buttonViewModels = [
+            StandardButtonViewModel(titleKey: "app_signInButton") {
```
Magdaline-A, line 24: How come we've been using the `StandardButtonViewModel` here instead of `AnalyticsButtonViewModel` ? @dylanjames-dd

Magdaline-A (approved):

Sources/Screens/Errors/DataDeletedWarningViewModel.swift
dylanjames-dd: This was previously an `AnalyticsButtonViewModel` but we reset the users analytics permissions before this screen is reached so it might be superfluous trying to send an analytics event when the user has reset their permissions. There could be an argument that we should send this and potentially allow firebase to send the event if they then restate their analytics permissions but there's been no requirement for that so the `StandardButtonViewModel` has therefore been used 🙂

dylanjames-dd (approved):
thank you!

Sources/Screens/Errors/DataDeletedWarningViewModel.swift
Magdaline-A: Ah, okay thanks!

PR #456, opened 2025-07-31
refactor: sign out flow

dylanjames-dd:
# refactor: sign out flow

Moving the sign out functionality to the SettingCoordinator and removing redundant code from the `resumeSession()` API.

# Checklist

~- [ ] Met all of the acceptance criteria specified in the user story on Jira~

### DIFF
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -157,21 +157,13 @@ final class PersistentSessionManager: SessionManager {
     }
     
     func resumeSession() throws {
-        guard expiryDate != nil else {
-            throw PersistentSessionError.noSessionExists
-        }
-        
-        guard isSessionValid else {
-            throw TokenError.expired
-        }
-        
         guard hasNotRemovedLocalAuth else {
             throw PersistentSessionError.userRemovedLocalAuth
         }
         
         let keys = try storeKeyService.fetch()
         if let idToken = keys.idToken {
-            try user.send(IDTokenUserRepresentation(idToken: idToken))
+            user.send(try IDTokenUserRepresentation(idToken: idToken))
         } else {
             user.send(nil)
         }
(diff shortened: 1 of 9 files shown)


Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -157,21 +157,13 @@ final class PersistentSessionManager: SessionManager {
     }
     
     func resumeSession() throws {
-        guard expiryDate != nil else {
-            throw PersistentSessionError.noSessionExists
-        }
-        
-        guard isSessionValid else {
-            throw TokenError.expired
```
chionia-gds, line 165: is there a chance that the session would expire right after the user just re-authenticated so we would get here and then the session would be expired when we run this code

dylanjames-dd: Not sure I understand what you mean.
When the user opens the app or brings it to the foreground `evaluateUser()` is run on the `AppQualifyingService` which, if the `sessionState` is `.saved`, will then call `resumeSession()` on the `PeristentSessionManager`. Evaluating `expiryDate` and `isSessionValid` when calling resume session is superfluous as ithas already been done by evaluating the `sessionState`

Magdaline-A (approved):

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
chionia-gds: as in, if the timing lines up and evaluateUser has already been called and then the session expires right before resumeSession is called.

chionia-gds (approved):

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
dylanjames-dd: I mean if it's all a synchronous call stack I don't think the possibility of that is very high. Also we've got a 10 second buffer built into `isSessionValid` so I think that won't be an issue.

PR #726, opened 2026-03-11
fix: DCMAW-18966 refresh token rollback

chionia-gds:
# DCMAW-18966 refresh token rollback

**Issue**: If a user has a refresh token and it's expiry date saved in the secure store, and then we rollback to only using access tokens for that user, they keep getting logged out because we are checking their saved refresh token expiry date every time they come back to the app, even if they just logged in

_Potential Flow_
1. User logs in with a refresh token which Is saved in `saveLoginTokens` called in `startAuthSession`
2. We rollback the user to only use access tokens because there was an issue with refresh tokens
3. User times out (i.e. their refresh token is expired)
4. User has to sign in again but there’s no refresh token returned this time -> they sign in with an access token
5. Because we have a stored refresh token, a refresh exchange is performed and `saveLoginTokens` doesn't save a new refresh token expiry date but also doesn't delete the old one
6. User opens the app again and we check that the refresh token is expired and log them out 

**Solution**
On step 5 above, we make sure to delete the old refresh token expiry date if a refresh token is not returned

# Checklist


shukriahmed-dd (approved):
looks good!!!

qnoid:
Looking at [PersistentSessionManager.swift#L99](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L99), the `OLString.refreshTokenExpiry` affects the contract of `validTokensForRefreshExchange`

Specifically, in case there is no `expiryDate`, `validTokensForRefreshExchange` [returns nil](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L103).

This is called by the [NetworkingService](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Sources/Service/OneLoginNetworkClient.swift#L37), **in case of nil**, "No refresh token or id token or valid access token, user must reauthenticate" which posts a `.reauthenticationRequired` and throws a `RefreshTokenExchangeError.reauthenticationRequired` error.

With that in mind, I looked to see what test coverage the `NetworkingService` has for the `func makeAuthorizedRequest(scope: String,request: URLRequest) ` as it relates to the `validTokensForRefreshExchange`. There are two tests:

* [test_makeAuthorisedRequest_invalidAccessToken](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Tests/UnitTests/Service/NetworkingServiceTests.swift#L88)
* [test_makeAuthorizedRequest_invalidTokens](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Tests/UnitTests/Service/NetworkingServiceTests.swift#L88)

While these tests do provide testing for the `NetworkingService`, they use a `MockSessionManager` which unfortunately means that changes to the `PersistentSessionManager` like the ones in this PR, will not catch regressions. 

Adding the following test:
```
    @Test("Test makeAuthorisedRequest() where there is no refresh token saved in the encrypted store")
    func test_makeAuthorisedRequest_when() async throws {
        
        let mockAccessControlEncryptedStore = MockSecureStoreService()
        let mockEncryptedStore = MockSecureStoreService()
        let mockUnprotectedStore = MockDefaultsStore()
        let mockLocalAuthentication = MockLocalAuthManager()

        let mockSessionManager = PersistentSessionManager(
            accessControlEncryptedStore: mockAccessControlEncryptedStore,
            encryptedStore: mockEncryptedStore,
            unprotectedStore: mockUnprotectedStore,
            localAuthentication: mockLocalAuthentication
        )
        
        let sut = NetworkingService(
            networkClient: NetworkClient(),
            refreshExchangeManager: MockRefreshTokenExchangeManager(),
            sessionManager: mockSessionManager
        )
        
        await #expect(throws: Never.self) {
            try await sut.makeAuthorizedRequest(
                scope: "",
                request: URLRequest(url: URL(string: "testurl.com")!)
            )
        }
```

Fails with a an error.

Screenshot 2026-03-12 at 15 40 33

What is the expected behavior of the app in this case? Do we expect the app to throw a `RefreshTokenExchangeError.reauthenticationRequired` error and post a `.reauthenticationRequired` notification in case a refresh token is no longer available?

chionia-gds:
@qnoid yes if there isn't any valid refresh tokens then the user needs to reauthenticate because then they have logged in using an access token and that is expired and needs to be refreshed. 

So if they logged in with a refresh token and then we rolled back and deleted the expiry date of the refresh token from the secure store, if they try to make an authorised request and their access token has expired they would need to re-authenticate because they no longer have a refresh token.

qnoid:
> @qnoid yes if there isn't any valid refresh tokens then the user needs to reauthenticate because then they have logged in using an access token and that is expired and needs to be refreshed.
> 
> So if they logged in with a refresh token and then we rolled back and deleted the expiry date of the refresh token from the secure store, if they try to make an authorised request and their access token has expired they would need to re-authenticate because they no longer have a refresh token.

Is that how we expect the app to work or indeed how the app works?

In that is indeed case, let's add [the test](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/726#issuecomment-4047827283)  and expect a `.reauthenticationRequired` error to indicate this is the expected behavior and prevent future regressions.

I would like some more time to review the code including any test coverage on the [PersistenceSessionManager](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift).

qnoid:
> if they try to make an authorised request and their access token has expired they would need to re-authenticate because they no longer have a refresh token.

[I can see that this is indeed how the app works](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/fix/dcmaw-18966-refresh-token-rollback/Sources/Service/OneLoginNetworkClient.swift#L36). The code does check first if an access token is valid; if not then checks to see if a refresh token is available to perform the refresh. This also seems to be the only place where the call to `validTokensForRefreshExchange` is made.

I am happy to approve once the test is added!

qnoid (approved):

SmithaGeetha-hub:
@chionia-gds Could you please add video evidence of the unit tests in the PR so QE can verify and close this ticket and make it ready for release?

We are expecting evidence on the following behaviours:

Refresh token issued – app behaving according to refresh token environment

Refresh token rolled back – app should behave based only on the access token, even if the previous refresh token has expired.

Refresh token enabled later – app should start behaving according to the refresh token environment again

This will help QE verify that all scenarios are covered. Thanks!
cc @jamielutton-gds

PR #774, opened 2026-04-13
feat: DCMAW-18889 update welsh content

shukriahmed-dd:
# feat: DCMAW-18889 update welsh content

# Checklist

### DIFF
--- a/Sources/Resources/cy-GB.lproj/Localizable.strings
+++ b/Sources/Resources/cy-GB.lproj/Localizable.strings
@@ -75,7 +75,7 @@
 
 
 // MARK: Generic error screen
-"app_genericErrorPage" = "Mae'n ddrwg gennym, mae yna broblem";
+"app_genericErrorPage" = "Mae'n ddrwg gennym, mae problem";
 
 "app_genericErrorPageBody" = "Rhowch gynnig arall yn nes ymlaen.";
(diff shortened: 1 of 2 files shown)


qnoid (approved):

qnoid:
🤔 should there have been an updated snapshot too? How come there is no snapshot test failing?

shukriahmed-dd:
@qnoid great question! Our snapshat tests only run in english so the update of the welsh doesn't cause tests to fail. I think this is fine since we have tests for our strings, and snapshot purpose is to see the screen with different config

qnoid:
> @qnoid great question! Our snapshat tests only run in english so the update of the welsh doesn't cause tests to fail. I think this is fine since we have tests for our strings, and snapshot purpose is to see the screen with different config

I see. That explains it. 


As a side note, I think it's worth having snapshot testing in Welsh. That way we can spot any layout issues in the Welsh language. It shouldn't take much to configure, probably just [adding Welsh as a language in the test plan configuration.](https://developer.apple.com/documentation/xcode/organizing-tests-to-improve-feedback#Adjust-configurations-for-test-plans).

alex-bradbury-gds (approved):

krissbennett-dd (approved):

PR #747, opened 2026-03-25
DCMAW-18499 improve reliability of UI tests

qnoid:
# [Improve reliability of UI Tests](https://govukverify.atlassian.net/browse/DCMAW-18499)

This PR ensures that UI Tests run as expected before expanding on their test coverage.
The proposed changes in this PR improve the reliability of UI Tests and in case of a failure, reduce the time it takes to report on it. Specifically, how long the [Manual UI Test](https://github.com/govuk-one-login/mobile-ios-one-login-app/actions/workflows/nightly-ui-test.yml) workflow runs.

As an example, the last run took 14 minutes to report on the UI Tests failing.
Screenshot 2026-03-25 at 14 47 26

As an example[^1] here is a failed run with the proposed code changes
image

[^1]: It's worth pointing out that this is not a direct comparison since the two environments (the Github runner and my laptop) are not identical but the factor between 4 and 14 minutes should give a good indication of the improvements we should expect to see.

# Changes

**This PR does not make any code changes to the application code**

At a high level, here are the changes for the **UITests**:

* Stop execution after failure so that the test process fails as soon as possible to avoid running in excess
* Deterministically wait for the app to be running on the foreground rather than at an arbitary time limit
    * ensure that each test uses its own instance of the `XCUIApplication` to operate to provide stronger guarantees on the state it operates
* Add assertions on wait for existence calls so as to fail early
* Add wait for the lock screen to disappear as a GIVEN for every test
* Add wait for existence calls where required (i.e. waiting for a *specific* element to appear on screen, say a tabbar for home, rather than a broader one like "any" view)

## test_loginHappyPath

This test scenario would sometimes fail because the "Loading" label does "exist" in the UI test. Looking at the video of a failed UI Test and by reviewing the application code, such case occurs where the "Loading" view controller is replaced **before** it can become available as a `XCUIElement`.

## test_loginCancelPath

This test scenario would sometimes fail because the assertion on the scroll view was on whether or not it was "hittable". Instead will now assert that the signInButton *exists* as a way to verify that the sign in screen is shown when cancel is tapped. This is a more reliable way than a scroll view being "hittable" as it is more lenient than requiring the view to exist *and* can be clicked, tapped, or pressed at its current location.

# Testing
Screenshot 2026-03-25 at 18 43 47

## Streak of runs
1. [Manual UI Test #135](https://github.com/govuk-one-login/mobile-ios-one-login-app/actions/runs/23553118518) ✅
[...]

PR #742, opened 2026-03-23
feat: DCMAW-18935 optimise running time for tests

qnoid:
# Worklflow optimisation
This PR optimises the workflow that runs when raising a PR, [which is now expected to run both unit and snapshot tests](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/740), to improve on its runtime and save costs.

**Note:** @Oliver-Binns brought to my attention [that public repositories on the standard GitHub-hosted runners is free](https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions) so the cost savings will only apply if we upgrade to the larger runners. 

## Changes

This PR changes the `pull-request` workflow to build the app and prepare it for testing. Effectively this allows for the app to be built once and reuse the build artifact to subsequently run both unit and snapshot tests.

At a high level this is the changeset for this PR:
* Modify the pull-request workflow to build, run unit tests, run snapshot tests in a single runner.
* Delete the snapshot test workflow since it's not longer used
* Modify the unit-test workflow to remove the dead code as it relates to the unit tests run as part of a PR 
* Add new fast lanes:
* * `buildForTesting`: which compiles the app to run tests
* * `testWithoutBuildingWithoutCoverage`: identical to the `testWithoutCoverage` used to run snapshot tests only without  building the app
* * `testWithoutBuilding`: identical to `test` used to run unit tests only without building the app

## Performance results

Before, we used 2 workflows to run unit tests and snapshot tests. Each workflow is assigned its own runner thus the 2 are running in parallel. Even though this setup allows for a **faster status report** on the PR, it is wasteful since the app is built twice, once for each set of tests to run on the respective runner.  

| workflow | time | status |
|:--------:|:--------:| :--------:|
| Unit Tests | 15m 15s | ✅ |
| Snapshot Tests | 12m 37s |  |
|  **Total Time**|  **27m 52s**|  |

Screenshot 2026-03-24 at 09 53 15

With the proposed changes, we use a single workflow to build the app, using the build artifact to subsequently run unit and snapshot tests.

| workflow | time | status |
|:--------:|:--------:| :--------:|
| Unit Tests + Snapshot Tests | 17m 28s | ✅ |
|  **Total Time**|  **17m 28s**|  |

Screenshot 2026-03-24 at 10 17 35

Even though that increases the time it takes to report on the status, the total runtime is down because of the savings made by not building the app twice. **This saves money**.

[...]

PR #777, opened 2026-04-14
feat: DMCAW-19321 reduce the time it takes to run unit tests

qnoid:
# [Reduce the time it takes to run unit tests](https://govukverify.atlassian.net/browse/DCMAW-19321)

| _ | execution time | 
|:--------:|:--------:|
| Before | 3m 08s |
| After | 1m 13s | 

**Before** 

before

**After**

after

# Changes

* Introduce a TestDelegate when running tests. The `TestDelegate` is conditional on an environment variable (i.e. IS_RUNNING_TESTS) which is expected to be set by a test plan when running logic tests. In the future, the same environment variable can be used when running UI Tests.

This is done in order to speedup test execution, eliminate any side effects and improve reliability. As an example, the `AppDelegate` configures Firebase which may have affected the [reliability of our UI Tests](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/747). The `SceneDelegate` initiates the `AppQualifyingService` which in turn performs a firebaseAppCheck under certain conditions.)

I have performed a first-time installation of the app on the simulator to test that the application delegate is correctly assigned and the scene is connected. Bellow is a screenshot of the app running for the first time using the main.swift file introduced in the PR.

Simulator Screenshot - iPhone 17 Pro - 2026-04-14 at 20 25 09

The change required replacing the [main attribute](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#main) in favor of a `main.swift` file that, at runtime evaluates the app delegate to use.

* Another change is the call to `FirebaseAppIntegrityService.firebaseAppCheck()`.  A reference to the `FirebaseAppIntegrityService.firebaseAppCheck()` function is passed on as a dependency with a default value of `FirebaseAppIntegrityService.firebaseAppCheck()` so that tests can provide a stub since Firebase is not configured in the `TestAppDelegate` when running tests like in the `AppDelegate`.

### Before
before

### After
after

This PR also resolves an issue where a log entry reported
> Class _TtC10Networking13NetworkClient is implemented in both 
	MockNetworking_2571C89223E3A5C0_PackageProduct.framework and OneLogin.app/OneLogin.debug.dylib 

with a warning of 

> "This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed." 

This was resolved by removing the MockNetworking library from the test target.

# Where to focus the review

* Please review the changes in the `AppQualifyingService` and `NetworkingService` as they relate to the `firebaseAppCheck()`. Does it work as expected?
 * Review the logic in the `main.swift` that assigns the correct delegate. Do you think there is a different implementation that minimises the risk of picking up the wrong delegate under the normal lifecycle of the app?

# Heads up

~This PR also [introduces a step](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/777/changes/1ac60670b0d3fb66f0557f3f8fb59420926b1e01) to upload the `OneLoginBuild.xcresult` for both unit and snapshot test runs.~ This has been updated [in another PR](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/779/) that has since been merged. This allows you to download the test results for inspection using Xcode. It should provide more clarity in figuring out which tests have failed as well as provide a tangible report that can be shared for visibility.

Screenshot 2026-04-15 at 21 21 20

# Checklist


Sources/Application/TestAppDelegate.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    private func removeCachedScenes(from application: UIApplication) {
+        /* Remove any cached scene configurations to ensure that
+         TestingAppDelegate.application(_:configurationForConnecting:options:) is called
+         and TestingSceneDelegate will be used when running unit tests.
+         
+         Warning: This is a private API and may break in the future.
+         It is perfectly app-store safe since it is not included in the main app targets. */
+        for sceneSession in application.openSessions {
+            application.perform(Selector(("_removeSessionFromSessionSet:")), with: sceneSession)
```
shukriahmed-dd, line 27: reading the documentation for https://developer.apple.com/documentation/objectivec/nsobjectprotocol/perform(_:) if this api does 'break' - what are the risks? will it stop us from being able to build the app?

shukriahmed-dd (commented):
For the changes to the tests, are they needed for this the main change to use a testing delegate or are these to reduce the test time further? 

If the latter is it possible to move these changes to a separate PR - sorry its just hard to see what changes are needed to accomodate the delegate change, thanks !

Sources/Application/main.swift
```diff
@@ -0,0 +1,13 @@
+import UIKit
+
+let isRunningTests = ProcessInfo.processInfo.environment["IS_RUNNING_TESTS"] == "1"
```
shukriahmed-dd, line 3: Is there a test that would fail if this property is set when the app is being built not for testing

qnoid:
> For the changes to the tests, are they needed for this the main change to use a testing delegate or are these to reduce the test time further?

These are to 
* remove any side effects,
* improve reliability
* reduce the test time further
mainly due to configuring and kickstarting Firebase. See [Failure waitForUnlockScreenNonExistence](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/747).


> If the latter is it possible to move these changes to a separate PR - sorry its just hard to see what changes are needed to accommodate the delegate change, thanks !

This [commit](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/777/changes/cb09405608854cd2b905c641d746832c5a4bf6e6) holds only the changes needed to accommodate the delegate change.

Sources/Application/TestAppDelegate.swift
qnoid: I don't expect it to prevent the app from being built.

The `_removeSessionFromSessionSet` is private APi. Not the `perform(_:)`. You can think of 
```
application.perform(Selector(("_removeSessionFromSessionSet:")), with: sceneSession)
```

As a Swift equivalent to 
```
application._removeSessionFromSessionSet(with: sceneSession)
```

Because the `_removeSessionFromSessionSet` function is **private**, you can use `perform(_:)` to call it.

The call to `_removeSessionFromSessionSet` is what allows the `TestAppDelegate` to set the `TestingSceneDelegate`.

If that method is renamed or its signature changes, effectively the call to [application(_:, configurationForConnecting: , options:)](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/777/changes#diff-88973806d32dcbac2ce5c7b5e5057b4fcb33f5cbc1991813410b8cdeb16ec49bR13), will not be made. 

Instead, the `TestAppDelegate` will use the `SceneDelegate`.

Effectively when running tests, instead of using the `TestingSceneDelegate`, it will revert back to the `SceneDelegate`.

Sources/Application/main.swift
qnoid: By property, do you mean the environment variable "IS_RUNNING_TESTS"?

If I understand correctly, you are proposing a test that will fail in case the app is expected to be built as normal but instead is built as if it's under test, am I correct? I do think there is merit in making sure the risk in wiring the application with the wrong delegate is as low as possible. I think that would be a good test to have, not sure how it would be possible to write a "unit" test against an application that is built NOT for testing. 🤔

In other words, **how can you test** that "the app is being built NOT for testing" **when the app is not under test**?

Even if you used an [assertionFailure](https://developer.apple.com/documentation/swift/assertionfailure(_:file:line:)) to stop execution, it's not immediately apparent to me how you could differentiate between a `IS_RUNNING_TESTS` environment variable set that is being expected vs not. I imagine you would have to have something else to differentiate on.

Sources/Application/AppDelegate.swift
```diff
@@ -2,7 +2,6 @@ import AppIntegrity
 import GAnalytics
 import UIKit
 
-@main
```
chionia-gds, line 5: why have we removed this?

qnoid: Every app must have a single main entry point.  The [main attribute](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#main) indicates that the annotated type, in this case the AppDelegate, provide a main type function as the entry point from which the application will start.

The changes in this PR instead use a `main.swift` file as the application entry point which allows us to write the code that at runtime checks the presence of the `IS_RUNNING_TESTS` environment variable to decide what delegate to assign to the `UIApplication`.

qnoid:
Will now track these changes as "tech debt" in two separate PRs
* https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/791
* https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/792

PR #783, opened 2026-04-20
fix: DCMAW-19786 a case of deeplink not handled once local auth has been successful

qnoid:
# [Fixed a case where the "deep link" is not handled](https://govukverify.atlassian.net/browse/DCMAW-19786)

Under certain conditions, the `WalletCoordinator` used to handle a "deep link" becomes nil. This PR ensures that the `WalletCoordinator` is never `nil` in that specific case.

Evidence of defect 

https://github.com/user-attachments/assets/42f6a2b8-8e83-4c59-bddd-dd6972c886e6

Evidence of fix

https://github.com/user-attachments/assets/8b9261aa-a90d-4420-a682-41cf0332cc9a

# Preface

The `WalletCoordinator` is an optional, [computed property](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/TabManagerCoordinator.swift#L32) in the [TabManagerCoordinator](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/TabManagerCoordinator.swift#L57) from the array of `childCoordinators`.

The [Coordination](https://github.com/govuk-one-login/mobile-ios-coordination) package, includes types like the [ParentCoordinator](https://github.com/govuk-one-login/mobile-ios-coordination/blob/1.3.0/Sources/Coordination/ParentCoordinator.swift), which the `TabManagerCoordinator` adopts and the [ChildCoordinator](https://github.com/govuk-one-login/mobile-ios-coordination/blob/1.3.0/Sources/Coordination/ChildCoordinator.swift#L7), which the WalletCoordinator adopts and makes it a child of the `TabManagerCoordinator`.

The `ChildCoordinator` defines a `finish` function. The [default implementation](https://github.com/govuk-one-login/mobile-ios-coordination/blob/1.3.0/Sources/Coordination/ChildCoordinator.swift#L20) and part of its contract, is to call `childDidFinish(_:)` on [its parent](https://github.com/govuk-one-login/mobile-ios-coordination/blob/1.3.0/Sources/Coordination/ParentCoordinator.swift#L35). On the condition that the child is indeed present in the array of `childCordinators`, the child is removed from the array.

So the question becomes, is there a use case where the `WalletCoordinator` has it `finish()` function called?

Here is the object graph for the `WalletCoordinator`.

* [WalletCoordinator](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Tabs/TabManagerCoordinator.swift#L85)
* * [walletAuthService](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/WalletCoordinator.swift#L24): [LocalAuthServiceWallet](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Utilities/WalletUtilities/LocalAuthServiceWallet.swift)
* * * [localAuthManager](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/WalletUtilities/LocalAuthServiceWallet.swift#L30): [OneLoginEnrolmentManager](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/OneLoginEnrolmentManager.swift#L24)
* * * * [coordinator](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/OneLoginEnrolmentManager.swift#L28): WalletCoordinator

The `OneLoginEnrolmentManager` type calls `finish` on the WalletCoordinator under the [completeEnrolment(isWalletEnrolment: completion:)](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/OneLoginEnrolmentManager.swift#L66)

This stack trace for this call is In case of a successful [save session](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/OneLoginEnrolmentManager.swift#L45), either in the case of:

* [A passcode is accepted as the only type available for local auth](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Utilities/WalletUtilities/LocalAuthServiceWallet.swift#L64).
* [Biometrics are accepted](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Utilities/WalletUtilities/LocalAuthServiceWallet.swift#L118).

**On an unrelated note** but I think it's worth pointing out that the there is another stack trace for the call in case of a successful [save session](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/OneLoginEnrolmentManager.swift#L45), which instead uses the [**EnrolmentCoordinator**](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Login/EnrolmentCoordinator.swift).

The stack trace for this is either in the case of:

* [A passcode](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Login/EnrolmentCoordinator.swift#L53)
* Biometrics [are allowed](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Login/EnrolmentCoordinator.swift#L46) or [skipped](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Login/EnrolmentCoordinator.swift#L48).

test_biometricsEnrolmentScreen_faceID 1

# Changes

The proposed code change is to only call finish when the coordinator is NOT a `WalletCoordinator`.

## Where to focus the review

Is this the expected behaviour of the code? Do you see any risk of a regression?

## Testing

Wrote 2 tests that demonstrate the defect and fail

* test_saveSession_isWalletEnrolmentTrue_finishOnCoordinator_not_called
* test_saveSession_isWalletEnrolmentTrue_walletCoordinator_notRemoved_asChild

Screenshot 2026-04-23 at 10 14 13

Applied the recommended fix and both tests now pass

Screenshot 2026-04-23 at 10 12 00

It's not clear how the code is expected to work. Which makes it hard to decide what test to write against.

The `Coordination` package suggests that every child coordinator has a lifecycle and that lifecycle has to finish at some point with the removal of the child from its parent. On the other hand, the `TabManagerCoordinator` expects the `WalletCoordinator` child to match its lifecycle. 

As of today, based on the product requirement that handling a deep link should be done for the duration of the lifecycle of the `TabManagerCoordinator`, it's safe to say this is the expected behavior.

However, should a new requirement arise in the future that expects the call to `WalletCoordinator.finish()` to have been made, any future code carries a risk of being blindsided by this fix at worse or have to grapple with the same decision at best.

Although I think this change carries a very low risk of regressions today for two reasons:
1. The proposed code change modifies a small part of a public contract (i.e. when/if the call to finish is made)
2. The affected code path is very narrow (i.e. to remove the wallet coordinator from the tab manager coordinator)

I would still like to use this PR as an opportunity to discuss and decide on the expected behaviour so a test can be written alongside the proposed code changes.

# Discussion

AFAICT, the call to `finish` on a child coordinator is done for two reasons:
* Memory management, i.e. release the coordinator once the work is done
* The work is not repeated by the same child coordinator once it's done.

Looking at the WalletCoordinator, it looks like it has two main responsibilities outside those enforced by any protocol adoption.
* [Handle the deep link](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/WalletCoordinator.swift#L82)
* [User cancellation](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/WalletCoordinator.swift#L87)

Since a defect has been raised due to not handling a deep link, (i.e. we expect the work to be repeatable) I think it's safe to assume that the `WalletCoordinator` should not be removed in this use case.

Which leaves the memory management side to it. With the proposed changes, the call to finish on the `WalletCoordinator` is never done. Which means the `WalletCoordinator` will be released once the `childCoordinators` array is released, which will be done when the `TabManagerCoordinator` is released, which seems to be when its call to `finish` is made, which is done when the [SettingsCoordinator is removed](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/SettingsCoordinator.swift#L89) and [a logout is posted](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Tabs/TabManagerCoordinator.swift#L124). That looks sufficient but happy to hear anything to the contrary.

```
extension TabManagerCoordinator: ParentCoordinator {
    func performChildCleanup(child: ChildCoordinator) {
        if child is SettingsCoordinator {
            NotificationCenter.default.post(name: .userDidLogout)
            finish()
        }
    }
}
```

# Checklist

### DIFF
--- a/Sources/Utilities/OneLoginEnrolmentManager.swift
+++ b/Sources/Utilities/OneLoginEnrolmentManager.swift
@@ -61,8 +61,8 @@ struct OneLoginEnrolmentManager: EnrolmentManager {
     func completeEnrolment(isWalletEnrolment: Bool = false, completion: (() -> Void)? = nil) {
         if !isWalletEnrolment {
             NotificationCenter.default.post(name: .enrolmentComplete)
+            coordinator?.finish()
         }
         completion?()
-        coordinator?.finish()
     }
 }
--- a/Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
+++ b/Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
@@ -1,33 +1,8 @@
 import Coordination
 import LocalAuthenticationWrapper
 import Networking
-import XCTest
-
 @testable import OneLogin
-
-class MockChildCoordinatorExpectation: ChildCoordinator {
-
-    weak var parentCoordinator: (any Coordination.ParentCoordinator)?
-
-    typealias StartAsFunction = () -> Void
-    typealias FinishAsFunction = () -> Void
-
-    var startAsFunction: StartAsFunction
-    var finishAsFunction: FinishAsFunction
-
-    init(startAsFunction: @escaping StartAsFunction = {}, finishAsFunction: @escaping FinishAsFunction = {}) {
-        self.startAsFunction = startAsFunction
-        self.finishAsFunction = finishAsFunction
-    }
-
-    func start() {
-        self.startAsFunction()
-    }
-
-    func finish() {
-        self.finishAsFunction()
-    }
-}
+import XCTest
 
 extension OneLoginEnrolmentManager {
     static func make(
@@ -51,6 +26,8 @@ extension OneLoginEnrolmentManager {
         )
     }
 }
+
+@MainActor
 final class OneLoginEnrolmentManagerTests: XCTestCase {
     private var mockLocalAuthContext: MockLocalAuthManager!
     private var mockSessionManager: MockSessionManager!
@@ -58,7 +35,6 @@ final class OneLoginEnrolmentManagerTests: XCTestCase {
     private var coordinator: ChildCoordinator!
     private var sut: OneLoginEnrolmentManager!
 
-    @MainActor
     override func setUp() {
         mockLocalAuthContext = MockLocalAuthManager()
         mockSessionManager = MockSessionManager()
@@ -90,7 +66,6 @@ final class OneLoginEnrolmentManagerTests: XCTestCase {
 }
 
 extension OneLoginEnrolmentManagerTests {
-    @MainActor
     func test_saveSession_succeeds() async {
         let exp = XCTNSNotificationExpectation(
             name: .enrolmentComplete,
@@ -105,7 +80,6 @@ extension OneLoginEnrolmentManagerTests {
         await fulfillment(of: [exp], timeout: 5)
     }
 
-    @MainActor
     func test_saveSession_fails() {
         // GIVEN the user has given FaceID permission
         mockLocalAuthContext.userDidConsentToFaceID = true
@@ -118,7 +92,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(mockAnalyticsService.crashesLogged, [MockError.generic as NSError])
     }
 
-    @MainActor
     func test_saveSession_promptForPermission_false() {
         // GIVEN the user has already given FaceID permission
         mockLocalAuthContext.userDidConsentToFaceID = false
@@ -129,7 +102,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(mockAnalyticsService.crashesLogged, [])
     }
 
-    @MainActor
     func test_saveSession_promptForPermission_cancelled() {
         // GIVEN promptForPermission throws a cancelled error
         mockLocalAuthContext.errorFromEnrolLocalAuth = LocalAuthenticationWrapperError.cancelled
@@ -140,7 +112,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(mockAnalyticsService.crashesLogged, [])
     }
 
-    @MainActor
     func test_saveSession_promptForPermission_fails() {
         // GIVEN promptForPermission throws an uncaught error
         mockLocalAuthContext.errorFromEnrolLocalAuth = MockError.generic
@@ -151,7 +122,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(mockAnalyticsService.crashesLogged, [MockError.generic as NSError])
     }
 
-    @MainActor
     func test_saveSession_isWalletEnrolmentTrue_finishOnCoordinator_not_called() {
         //  GIVEN OneLoginEnrolmentManager with a coordinator
         //  WHEN performing save session
@@ -171,7 +141,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(result, .completed)
     }
 
-    @MainActor
     func test_saveSession_isWalletEnrolmentFalse_finishOnCoordinator_called() {
         //  GIVEN OneLoginEnrolmentManager with a coordinator
         //  WHEN performing save session
@@ -190,7 +159,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(result, .completed)
     }
 
-    @MainActor
     func test_saveSession_default_finishOnCoordinator_called() {
         //  GIVEN OneLoginEnrolmentManager with a coordinator
         //  WHEN performing save session (where by default `isWalletEnrolment` is false)
@@ -208,7 +176,6 @@ extension OneLoginEnrolmentManagerTests {
         XCTAssertEqual(result, .completed)
     }
 
-    @MainActor
     func test_saveSession_isWalletEnrolmentTrue_walletCoordinator_notRemoved_asChild() {
         //  GIVEN a `TabManagerCoordinator`
         //  AND a `WalletCoordinator`
--- a/Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
+++ b/Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
@@ -147,7 +147,7 @@ extension OneLoginEnrolmentManagerTests {
     }
 
     @MainActor
-    func test_saveSession_isWalletEnrolmentTrue_callsFinishOnCoordinator() {
+    func test_saveSession_isWalletEnrolmentTrue_finishOnCoordinator_not_called() {
         let expectation = expectation(description: #function)
         expectation.isInverted = true
         let mockChildCoordinatorExpectation = MockChildCoordinatorExpectation(finishAsFunction: {
@@ -163,7 +163,7 @@ extension OneLoginEnrolmentManagerTests {
     }
     
     @MainActor
-    func test_saveSession_isWalletEnrolmentFalse_finishOnCoordinatorNotCalled() {
+    func test_saveSession_isWalletEnrolmentFalse_finishOnCoordinator_called() {
         let expectation = expectation(description: #function)
         let mockChildCoordinatorExpectation = MockChildCoordinatorExpectation(finishAsFunction: {
                 expectation.fulfill()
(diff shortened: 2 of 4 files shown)


qnoid:
Now tracking it as https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/785

Sources/Utilities/OneLoginEnrolmentManager.swift
```diff
@@ -63,6 +63,10 @@ struct OneLoginEnrolmentManager: EnrolmentManager {
             NotificationCenter.default.post(name: .enrolmentComplete)
         }
         completion?()
+        
+        guard !(coordinator is WalletCoordinator) else {
+            return
+        }
         coordinator?.finish()
```
robinphillips: I see this fixes the problem by not finishing wallet coordinator. From what I can see the completeEnrolment function is only called if the coordinator is WalletCoordinator . I wonder if it might be better to simply remove the `finish()` call instead?

Sources/Utilities/OneLoginEnrolmentManager.swift
```diff
@@ -63,6 +63,10 @@ struct OneLoginEnrolmentManager: EnrolmentManager {
             NotificationCenter.default.post(name: .enrolmentComplete)
         }
         completion?()
+        
+        guard !(coordinator is WalletCoordinator) else {
+            return
```
shukriahmed-dd: what about putting the `coordinator?.finish()` inside the if block instead of returning early 

```
if !isWalletEnrolment {
            NotificationCenter.default.post(name: .enrolmentComplete)
            coordinator?.finish()
}
```

qnoid: I am happy to make [the change](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/783/changes#diff-c9b761ba3077f5c19b49f8f4cdeddb62b8a72c5dcb0c83d8fa7bbba322315bc6R64).

The main drawback of using the `isWalletEnrolment` flag to decide whether to call finish or not is that it **requires** the calling code that uses the `WalletCoordinator` as a dependency to the `OneLoginEnrolmentManager` to also pass in `false` to the `isWalletEnrolment`.

This creates a coupling between the `WalletCoordinator` and the `isWalletEnrolment` that isn't obvious to the calling code. Which leaves room for the defect to resurface in the case where `isWalletEnrolment` is false while still using a `WalletCoordinator`.

To be fair, even the proposed code change in this PR that ensures the `finish` is not called on the `WalletCoordinator`, assumes that the `TabManagerCoordinator` is involved which goes against its own invariants of requiring the `WalletCoordinator` to never be removed from the array of `childCoordinators` for its lifecycle. 

This might cause an issue in the future because: 

> However, should a new requirement arise in the future that expects the call to WalletCoordinator.finish() to have been made, any future code carries a risk of being blindsided by this fix at worse or have to grapple with the same decision at best.

Ideally, I think the decision on whether or not to call `finish` should be the responsibility of the `LocalAuthServiceWallet`, for the `WalletCoordinator` and the `EnrolmentCoordinator` in the `completion` block that the `OneLoginEnrolmentManager.saveSession()` accepts.

Finally, generally speaking this part of the codebase tightly couples the `WalletCoordinator` to the `TabManagerCoordinator` to the `OneLoginEnrolmentManager` that leaves room for defects like this one. In the future we should aim to either decouple or provide an API that leaves less room for misuse.

Sources/Utilities/OneLoginEnrolmentManager.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/785#pullrequestreview-4153411917

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
```diff
@@ -102,4 +145,73 @@ extension OneLoginEnrolmentManagerTests {
         // THEN an error is recorded in Crashlytics
         XCTAssertEqual(mockAnalyticsService.crashesLogged, [MockError.generic as NSError])
     }
+
+    @MainActor
+    func test_saveSession_isWalletEnrolmentTrue_finishOnCoordinator_not_called() {
+        let expectation = expectation(description: #function)
+        expectation.isInverted = true
```
shukriahmed-dd, line 132: question, instead of `expectation.isInverted` is it possible to have the assertion be false?  this tests looks similar to the one below so it hard to spot that were not actually equaling `result` and `.completed` . If its not possible can we add a comment above this line please :)

shukriahmed-dd (commented):
Looks good! Some of the spacing looks off could we 'control I' the tests please :)

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
```diff
@@ -102,4 +145,73 @@ extension OneLoginEnrolmentManagerTests {
...
+                                        sessionManager: mockSessionManager)
+        
+        let walletCoordinator =  WalletCoordinator(analyticsService: mockAnalyticsService,
+                                               networkingService: NetworkClient(),
+                                               sessionManager: mockSessionManager)
+        
+        tabManagerCoordinator.childCoordinators.append(walletCoordinator)
+        walletCoordinator.parentCoordinator = tabManagerCoordinator
+
+
```
shukriahmed-dd: nit: extra line here

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
qnoid: Not sure what you mean by the "assertion to be false".  An `inverted` expectation is one where [you don't expect fulfill to be called](https://developer.apple.com/documentation/xctest/xctestexpectation/isinverted). Which is what this test is asserting. Should the expectation be fulfilled, i.e. finish is called, the test will fail.

I will add a GIVEN, WHEN, THEN comment above to explain what the test does.

shukriahmed-dd: ah yes I see, is it possible to do? 

`XCTAssertEqual(result.isInverted, .completed)`

qnoid: Nope.

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
```diff
@@ -1,15 +1,63 @@
 import Coordination
 import LocalAuthenticationWrapper
-@testable import OneLogin
+import Networking
 import XCTest
 
+@testable import OneLogin
```
shukriahmed-dd, line 4: can this be moved to be under the networking import, please

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
```diff
@@ -1,15 +1,63 @@
...
+
+class MockChildCoordinatorExpectation: ChildCoordinator {
```
shukriahmed-dd: sorry can this mock be moved to a separate file, i think theres a folder for mocks

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
shukriahmed-dd, line 68: can @ MainActor be added above this extension please and removed from on top of the functions - looks like they all need this

shukriahmed-dd (approved):

chionia-gds (approved):

PR #791, opened 2026-04-27
DMCAW-0000 use test delegate when the application is under testing

qnoid:
# Replace the AppDelegate with a TestDelegate

The proposed code changes in this PR aim to

* remove any side effects
* improve reliability
* reduce the time it takes to run tests 

as they relate mainly due to configuring and kickstarting Firebase which is what the AppDelgate does. 

# Changes

* Introduce a TestDelegate when running tests. The `TestDelegate` is conditional on an environment variable (i.e. IS_RUNNING_TESTS) which is expected to be set by a test plan when running logic tests. In the future, the same environment variable can be used when running UI Tests.

This is done in order to speedup test execution, eliminate any side effects and improve reliability. As an example, the `AppDelegate` configures Firebase which may have affected the [reliability of our UI Tests](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/747). The `SceneDelegate` initiates the `AppQualifyingService` which in turn performs a firebaseAppCheck under certain conditions.)

I have performed a first-time installation of the app on the simulator to test that the application delegate is correctly assigned and the scene is connected. Bellow is a screenshot of the app running for the first time using the main.swift file introduced in the PR.

Simulator Screenshot - iPhone 17 Pro - 2026-04-14 at 20 25 09

The change required replacing the [main attribute](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#main) in favor of a `main.swift` file that, at runtime evaluates the app delegate to use.

* Another change is the call to `FirebaseAppIntegrityService.firebaseAppCheck()`.  A reference to the `FirebaseAppIntegrityService.firebaseAppCheck()` function is passed on as a dependency with a default value of `FirebaseAppIntegrityService.firebaseAppCheck()` so that tests can provide a stub since Firebase is not configured in the `TestAppDelegate` when running tests like in the `AppDelegate`.

This PR also resolves an issue where a log entry reported
> Class _TtC10Networking13NetworkClient is implemented in both 
	MockNetworking_2571C89223E3A5C0_PackageProduct.framework and OneLogin.app/OneLogin.debug.dylib 

with a warning of 

> "This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed." 

This was resolved by removing the MockNetworking library from the test target.

# Where to focus the review

The vast majority of code changes do not change any of the test assertions, other than a few rare cases where it was deemed necessary.

* Please double check the changes on each test case to ensure no accidental changes to the tests have been made.
[...]

PR #792, opened 2026-04-27
DMCAW-0000 improve execution time on unit tests

qnoid:
# Reduce the time it takes to run unit tests

<blockquote>
**Note:**  This branch is based off `fix/DCMAW-0000-use-test-delegate-when-under-testing` in #791 . This can be merged once #791 has.
</blockquote>

| _ | execution time | 
|:--------:|:--------:|
| Before | 3m 08s |
| After | 1m 13s | 

**Before** 

before

**After**

after

# Changes

### Before
before

### After
after

# Where to focus the review

# Checklist

### DIFF
--- a/Tests/UnitTests/Login/LoginCoordinatorTests.swift
+++ b/Tests/UnitTests/Login/LoginCoordinatorTests.swift
@@ -6,6 +6,47 @@ import GDSCommon
 import SecureStore
 import XCTest
 
+class MockNavigationControllerExpectation: UINavigationController {
+    var expectation: XCTestExpectation?
+    
+    init(expectation: XCTestExpectation? = nil) {
+        self.expectation = expectation
+        super.init(nibName: nil, bundle: nil)
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+    }
+    
+    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        super.pushViewController(viewController, animated: animated)
+        self.expectation?.fulfill()
+    }
+}
+
+extension LoginCoordinator {
+    
+    static func make(mockNavigationController: UINavigationController = UINavigationController(), mockSessionManager: SessionManager = MockSessionManager()) -> LoginCoordinator {
+        
+        let appWindow = UIWindow()
+        let mockAnalyticsService = MockAnalyticsService()
+        let mockNetworkMonitor = MockNetworkMonitor()
+        let mockAuthenticationService = MockAuthenticationService(sessionManager: mockSessionManager)
+        appWindow.rootViewController = mockNavigationController
+        appWindow.makeKeyAndVisible()
+
+        return LoginCoordinator(appWindow: appWindow,
+                               root: mockNavigationController,
+                               analyticsService: mockAnalyticsService,
+                               sessionManager: mockSessionManager,
+                               networkMonitor: mockNetworkMonitor,
+                               authService: mockAuthenticationService,
+                               sessionState: .notLoggedIn,
+                               serviceState: nil)
+        
+    }
+}
+
 final class LoginCoordinatorTests: XCTestCase {
     var appWindow: UIWindow!
     var navigationController: UINavigationController!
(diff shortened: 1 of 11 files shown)


Tests/UnitTests/Login/LoginCoordinatorTests.swift
```diff
@@ -61,6 +84,66 @@ final class LoginCoordinatorTests: XCTestCase {
...
+
+        wait(for: [startAuthSessionExpectation, pushViewControllerExpectation], timeout: 10)
+        XCTAssertTrue(mockSessionManager.didCallStartSession)
+        
+        let vc = try XCTUnwrap(mockNavigationController.topViewController as? GDSErrorScreen)
+        
+        return vc.viewModel
+    }
+    
+    @MainActor
```
chionia-gds, line 114: same thing here to move @mainactor to the class type

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
chionia-gds, line 47: can't we use 
```
  let mockAnalyticsService = MockAnalyticsService()
        let sut: OneLoginEnrolmentManager = .make(mockLocalAuthContext: mockLocalAuthManagerExpectation,
                                                  mockSessionManager: mockSessionManager,
                                                  mockAnalyticsService: mockAnalyticsService)
```
here instead of re-writing this code on every test. 

it seems that now this sut isn't used

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
```diff
@@ -81,43 +81,81 @@ extension OneLoginEnrolmentManagerTests {
...
+        wait(for: [expectation], timeout: 5)
+        XCTAssertTrue(mockLocalAuthManager.didCallEnrolFaceIDIfAvailable)
         // THEN no error is recorded in Crashlytics
         XCTAssertEqual(mockAnalyticsService.crashesLogged, [])
     }
 
     func test_saveSession_promptForPermission_fails() {
+        let expectation = expectation(description: #function)
         // GIVEN promptForPermission throws an uncaught error
+        let mockLocalAuthContext = MockLocalAuthManager()
```
chionia-gds, line 149: this isn't needed here we set mockLocalAuthContext in the setUp()

chionia-gds: same holds for the test cases above where we re-set 
 let mockAnalyticsService = MockAnalyticsService()
let mockLocalAuthManager = MockLocalAuthManager()
 let mockSessionManager = MockSessionManager()
even though they are set in setup()

and all other tests in this PR

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
```diff
@@ -421,7 +472,8 @@ extension PersistentSessionManagerTests {
         )
         
         // THEN my session data is cleared
-        waitForTruth(self.didCall_deleteSessionBoundData, timeout: 5)
+        await fulfillment(of: [didCall_deleteSessionBoundData], timeout: 5.0)
```
chionia-gds, line 475: why are we mixing: 
wait(for: ) and await fulfillment(of:) 

Can't we use `await fulfillment(of:)` in all the test cases where we use `wait` in this PR?

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
qnoid: The reason for using a local variable is twofold:
* Not have to init/reset the shared state for each test in the setUp/tearDown
* Make it easy to switch running tests in parallel

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
qnoid: The main issue with using a shared instance of `OneLoginEnrolmentManager` across our tests is that there is a risk of interference between the tests which requires to both reset the state after each test has run:

```
        mockLocalAuthContext = nil
        mockSessionManager = nil
        mockAnalyticsService = nil
        coordinator = nil
        sut = nil
```

as well as forcing the test case to run serially on the `@MainActor`.

The `OneLoginEnrolmentManager.make()` function is meant to provide a an instance of `OneLoginEnrolmentManager` with reasonable defaults while also allowing each test to provide customisation (e.g. setting the state of a mock to be used)

Happy to introduce more `make()` functions that produce an instance of `OneLoginEnrolmentManager` configured for tests that want to use the same configuration. Do you see any such cases?

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
qnoid: Using a `wait(for:)` in an async function issues a warning; which in Swift 6 is an outright error.
 
> Instance method 'wait' is unavailable from asynchronous contexts; Use await fulfillment(of:timeout:enforceOrder:) instead; this is an error in the Swift 6 language mode

qnoid:
Will now track this PR at https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/796

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
chionia-gds: yeah so can't we use await fulfillment(of:) in all the instances then?

PR #796, opened 2026-04-29
DCMAW-20025 reduce the time it takes to run unit tests

qnoid:
# [Reduce the time it takes to run unit tests](https://govukverify.atlassian.net/browse/DCMAW-20025)

| _ | execution time | 
|:--------:|:--------:|
| Before | 3m 08s |
| After | 1m 13s | 

**Before** 

before

**After**

after

# Changes

### Before
before

### After
after

# Where to focus the review

# Checklist

### DIFF
--- a/Tests/UnitTests/Login/LoginCoordinatorTests.swift
+++ b/Tests/UnitTests/Login/LoginCoordinatorTests.swift
@@ -29,6 +29,7 @@ extension LoginCoordinator {
     }
 }
 
+@MainActor
 final class LoginCoordinatorTests: XCTestCase {
     var appWindow: UIWindow!
     var navigationController: UINavigationController!
@@ -38,7 +39,6 @@ final class LoginCoordinatorTests: XCTestCase {
     var mockAuthenticationService: MockAuthenticationService!
     var sut: LoginCoordinator!
     
-    @MainActor
     override func setUp() {
         super.setUp()
         
@@ -72,7 +72,6 @@ final class LoginCoordinatorTests: XCTestCase {
         super.tearDown()
     }
     
-    @MainActor
     func reauthLogin() {
         mockSessionManager.isReturningUser = true
         sut = LoginCoordinator(appWindow: appWindow,
@@ -85,7 +84,6 @@ final class LoginCoordinatorTests: XCTestCase {
                                serviceState: nil)
     }
     
-    @MainActor
     func given(errorFromStartSession: Error, when: (LoginCoordinator) -> Void) throws -> GDSErrorViewModelV3 {
         let startAuthSessionExpectation = expectation(description: #function)
         let pushViewControllerExpectation = self.expectation(description: #function)
@@ -111,7 +109,6 @@ final class LoginCoordinatorTests: XCTestCase {
         return vc.viewModel
     }
     
-    @MainActor
     func given(errorFromStartSession: Error,
                repeats count: Int,
                when: (LoginCoordinator) -> Void,
@@ -152,7 +149,7 @@ enum AuthenticationError: Error {
 
 extension LoginCoordinatorTests {
     // MARK: Login
-    @MainActor
+    
     func test_start() {
         // WHEN the LoginCoordinator is started
         sut.start()
@@ -161,7 +158,6 @@ extension LoginCoordinatorTests {
         XCTAssertTrue(sut.root.topViewController is IntroViewController)
     }
     
-    @MainActor
     func test_start_reauth() throws {
         // WHEN the LoginCoordinator is started in a reauth flow
         reauthLogin()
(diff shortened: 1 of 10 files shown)


Tests/UnitTests/Tabs/SettingsCoordinatorTests.swift
```diff
@@ -102,9 +133,13 @@ final class SettingsCoordinatorTests: XCTestCase {
         // WHEN the user signs out
         let signOutButton: UIButton = try XCTUnwrap(presentedVC.topViewController?.view[child: "instructions-button"])
         signOutButton.sendActions(for: .touchUpInside)
+        
+        wait(for: [pushViewControllerExpectation], timeout: 10)
```
chionia-gds, line 114: shouldn't we use await fulfilment for consistency?

qnoid: We cannot use `await` since the test function isn't async.

We could make a coordinated effort to migrate all of our tests to use await fulfillment but I think that is outside the scope of this PR.

chionia-gds (approved):

Tests/UnitTests/Login/LoginCoordinatorTests.swift
```diff
@@ -91,15 +170,22 @@ extension LoginCoordinatorTests {
...
     
-    @MainActor
     func test_authenticate_launchAuthenticationService() {
+        let expectation = expectation(description: #function)
+        let mockSessionManager = MockSessionManager()
+        let mockSessionManagerExpectation = MockSessionManagerExpectation(sessionManager: mockSessionManager, didStartAuthSessionAsFunction: { _, _ in
+            expectation.fulfill()
+        })
+        
+        let sut: LoginCoordinator = .make(mockSessionManager: mockSessionManagerExpectation)
```
shukriahmed-dd, line 141: With the factory functions being created in this PR, can all the tests create a local sut so the setUp and tearDown functions removed? From our chat our Friday seems like the goal is to use factory helps instead of using variables shared across tests

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
```diff
@@ -8,6 +8,35 @@ import MockNetworking
...
+    let expectation: XCTestExpectation
+        
+    func clearSessionData() {
+        self.expectation.fulfill()
+    }
+}
+
+extension PersistentSessionManager {
+    
+    static func make(mockEncryptedStore: MockSecureStoreService = MockSecureStoreService(),
```
shukriahmed-dd, line 21: based on our chat Friday, is it best PersistentSessionManagerTests style is not edited until we can do tech debt that changes all

qnoid: These two tests use expectations, unlike every other test. Either setup must be modified to conditionally setup the mocks with expectations or every other test has to be modified to work with expectations. That is to say, the changes are not as trivial.

Tests/UnitTests/Login/LoginCoordinatorTests.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/796/commits/be2159a584ac6c5b61d9ecb75fee1e5b285e8aaf

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
shukriahmed-dd: Happy for this to be tech debt done outside this ticket

Tests/UnitTests/Utilities/OneLoginEnrolmentManagerTests.swift
shukriahmed-dd, line 48: looks like we only have this test using setUp in this file, can we change this to create local sut and delete setUp and tearDown please :)

qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/796/changes/86cd1850ef8b8bac3c58ff241bffae7123a735fd

Tests/UnitTests/Tabs/SettingsCoordinatorTests.swift
shukriahmed-dd, line 130: sorry didn't notice this file before, could we do the same here and create local sut using make(), happy to approve after this :)

qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/796/commits/11857d9c41653a58f56c83417bbdafeebca30101

shukriahmed-dd (commented):
looks good! thanks for the changed :)

shukriahmed-dd (approved):

PR #797, opened 2026-04-29
DCMAW-20026 use test delegate when the application is under testing

qnoid:
# [Replace the AppDelegate with a TestDelegate](https://govukverify.atlassian.net/browse/DCMAW-20026)

The proposed code changes in this PR aim to

* remove any side effects
* improve reliability
* reduce the time it takes to run tests 

as they relate mainly due to configuring and kickstarting Firebase which is what the AppDelgate does. 

# Changes

* Introduce a TestDelegate when running tests. The `TestDelegate` is conditional on an environment variable (i.e. IS_RUNNING_TESTS) which is expected to be set by a test plan when running logic tests. In the future, the same environment variable can be used when running UI Tests.

This is done in order to speedup test execution, eliminate any side effects and improve reliability. As an example, the `AppDelegate` configures Firebase which may have affected the [reliability of our UI Tests](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/747). The `SceneDelegate` initiates the `AppQualifyingService` which in turn performs a firebaseAppCheck under certain conditions.)

I have performed a first-time installation of the app on the simulator to test that the application delegate is correctly assigned and the scene is connected. Bellow is a screenshot of the app running for the first time using the main.swift file introduced in the PR.

Simulator Screenshot - iPhone 17 Pro - 2026-04-14 at 20 25 09

The change required replacing the [main attribute](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#main) in favor of a `main.swift` file that, at runtime evaluates the app delegate to use.

* Another change is the call to `FirebaseAppIntegrityService.firebaseAppCheck()`.  A reference to the `FirebaseAppIntegrityService.firebaseAppCheck()` function is passed on as a dependency with a default value of `FirebaseAppIntegrityService.firebaseAppCheck()` so that tests can provide a stub since Firebase is not configured in the `TestAppDelegate` when running tests like in the `AppDelegate`.

This PR also resolves an issue where a log entry reported
> Class _TtC10Networking13NetworkClient is implemented in both 
	MockNetworking_2571C89223E3A5C0_PackageProduct.framework and OneLogin.app/OneLogin.debug.dylib 

with a warning of 

> "This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed." 

This was resolved by removing the MockNetworking library from the test target.

# Where to focus the review

* Please review the changes in the `AppQualifyingService` and `NetworkingService` as they relate to the `firebaseAppCheck()`. Does it work as expected?
 * Review the logic in the `main.swift` that assigns the correct delegate. Do you think there is a different implementation that minimises the risk of picking up the wrong delegate under the normal lifecycle of the app?

# Background reading
* [What does @main do in Swift 5.3?](https://useyourloaf.com/blog/what-does-@main-do-in-swift-5.3/)
* [UIApplicationMain(_:_:_:_:)](https://developer.apple.com/documentation/uikit/uiapplicationmain(_:_:_:_:)-1yub7)
* [Testing Tips & Tricks](https://developer.apple.com/videos/play/wwdc2018/417/)
* [Testing Tips & Tricks Presentation Slide, Page 170](https://devstreaming-cdn.apple.com/videos/wwdc/2018/417j8ucs9p8w7seip/417/417_testing_tips__tricks.pdf?dl=1)

Screenshot 2026-04-27 at 12 53 30

# Checklist

### DIFF
--- a/Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
+++ b/Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
@@ -1,4 +1,6 @@
 // swiftlint:disable file_length
+import AppIntegrity
+import GAnalytics
 import MobilePlatformServices
 import Networking
 @testable import OneLogin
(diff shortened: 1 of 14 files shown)


chionia-gds (approved):

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
```diff
@@ -1,418 +1,510 @@
...
-            timeout: 5
-        )
+        XCTAssertEqual(appState, .unavailable)
     }
 }
 
 // MARK: - User State Evaluation
 extension AppQualifyingServiceTests {
+    
+    func waitForSessionStateChange(expectation e: XCTestExpectation? = nil,
```
shukriahmed-dd: nit: can these two helper functions be moved to the bottom of the file, along with the class AppQualifyingServiceDelegateExpectation please

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
```diff
@@ -1,418 +1,509 @@
...
-
+        let sut: AppQualifyingService = .make(appInformationProvider: appInformationProvider)
+        
+        let (appState, _) = waitForAppInfoStateChange(sut: sut, when: { sut in
+            sut.initiate()
+        })
+        
+        XCTAssertEqual(appState, .qualified)
         XCTAssertEqual(AppEnvironment.remoteReleaseFlags.flags, releaseFlags)
     }
```
shukriahmed-dd, line 81: what if along with this test, theres a test checking the isRunningTests Bool e.g.

```
func test_isRunningTests_value() {
        #if DEBUG
        XCTAssertTrue(isRunningTests)
        #else
        XCTFail("This test should be ran on the Debug configuration")
        #endif
    }
```

qnoid: Can you explain a bit more as to what the value of this test would be?

shukriahmed-dd: To explicitly test isRunningTests value , so if its changed to 0 this fail as well as debug config being used for tests. Understand we might have tests that would fail in those cases, but thought it would be nice to have an explicit tests

qnoid: I am a bit reluctant to add a test that neither exercises the source code nor how the app is built.

The test you are describing effectively asserts that, **the test plan for which this specific test will be written and is part of** actually defines an environment variable to indicate it is indeed running tests. I am still unsure of the value it adds.

shukriahmed-dd: Since the environment variable is created in app code, shouldn't it be tested? If the following happens:
a. The environment variable is set to 0 
b. The testAppDelgate is used to build the app
Which tests will fail? I thought it would be good to have specific tests for these scenarios, instead of relying on other tests to fail/ not build

qnoid: I am afraid I don't quite follow. Can you please write the test as you envision it?

shukriahmed-dd: I haven't envisioned the tests for these scenarios :) Wondering with how the test our right now, what we would see happen for scenario a & b

qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/777#discussion_r3132133131

shukriahmed-dd (approved):
looks good !

PR #806, opened 2026-05-07
fix: DCMAW-19978 a case of a refresh token reuse

qnoid:
# [iOS | Refresh token reuse (sync anomaly)](https://govukverify.atlassian.net/browse/DCMAW-19978?atlOrigin=eyJpIjoiMGQzYjFhYTQzY2MxNDY0NWIxZGE1MjRiM2I5MmYzYWIiLCJwIjoiaiJ9)

This PR fixes a case where the **same** refresh token will be sent to perform a refresh exchange.

**Note:** this is an issue that has been found and reported by our logs. That is to say, as of this moment, there are no steps to reproduce this within the app. Instead, this PR is the product of a code review that indeed shows how re-using a refresh token is probable. Having said that, however unlikely, it is possible that there is another use case where the refresh token is being re-used that is not covered by the changes proposed in this PR.

# Preface

Both the `PersistentSessionManager` and `NetworkingService` use the `getUpdatedTokens`, which is the request for the refresh exhange, in the [resumeSession(tokenExchangeManager:appIntegrityProvider:)](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L276) and [makeAuthorizedRequest(scope:request:)](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.19.1/Sources/Service/NetworkingService.swift#L67) respectively.

Both functions, as they relate to the refresh token exchange, effectively:
* Read the refresh token from the store (`TokenStore`)
* Send a request to get a new value (`getUpdatedTokens`)
* Store the token for the future requests

The issue raised suggests that the code is re-using a refresh token when sending the request. Code review shows that the code related to the refresh token exchange is `async`. Effectively, there can be two threads (i.e. tasks), each reading the same token, one succeeds in sending a request to get a new value, the late one attempts to use the same refresh token at which point it fails.

AFAICT, this code can be called from the **OneLogin** app as well as the **WalletSDK** and **IDCheck**.

# Changes

This code change introduces a `SerialTaskQueue` type and modifies both the `NetworkingService` and `PersistenceSessionManager` to perform the refresh exchange atomically by using the queue.

In practice, the proposed code change modifies both the `resumeSession` and `makeAuthorizedRequest` function to instead  `enqueue` their operations, effectively making the read/update/write atomic. Effectively, wrapping both their implementations in a  closure like so:

```
try await self.serialTaskQueue.enqueue {

}
```

Since the sequence to the read/update/write the refresh token can happen on either and/or both, the `SerialTaskQueue` is shared so that the operation is atomic across the `resumeSession` and `makeAuthorizedRequest`.

# Testing

**Note:** These changes have not been tested by running the app or on any user scenarios.

I was able to write the following tests that replicate (i.e. a refresh token used twice) and report the issue. With the proposed code changes, the tests pass successfully.

## test_makeAuthorisedRequest_invalidAccessToken_concurrent
test_makeAuthorisedRequest_invalidAccessToken_concurrent fail
test_makeAuthorisedRequest_invalidAccessToken_concurrent success

## test_refreshTokenExchange_isSerialisedAcrossResumeSessionAndAuthorizedRequest

test_refreshTokenExchange_isSerialisedAcrossResumeSessionAndAuthorizedRequest fail
test_refreshTokenExchange_isSerialisedAcrossResumeSessionAndAuthorizedRequest success

## test_makeAuthorisedRequest_invalidAccessToken_concurrent_with_SessionManager

test_makeAuthorisedRequest_invalidAccessToken_concurrent_with_SessionManager fail
test_makeAuthorisedRequest_invalidAccessToken_concurrent_with_SessionManager success

# Where to focus the review

* Ensure the tests are correctly setup and exercise the use case as described in this PR.
* Ensure the proposed code change does not introduce any side effects.
* Ensure the `SerialTaskQueue` has enough and well written test coverage.
* Are there any other instances of a `PersistenceSessionManager` and `NetworkingService` that should be sharing a `SerialTaskQueue`?

# What this PR does not do

This PR does not otherwise alter the behavior of the code. e.g.
* The ordering in which the refresh token exchange request is made is still arbitrary between the `NetworkingService` and `PersistentSessionManager`.
* The code change does not reduce or otherwise prevent a refresh token exchange from being made unnecessarily (e.g. when a refresh token has just been exchanged).

Both are considered outside the scope of this PR and not necessary to fix the violation.

# Noteworthy additions

* `SerialTaskQueue`, A FIFO queue that guarantees a serialised execution of `async` operations.

# Checklist


Sources/Service/NetworkingService.swift
```diff
@@ -34,21 +37,23 @@ final class NetworkingService {
...
-                
-                return try await networkClient.makeAuthorizedRequest(
-                    scope: scope,
-                    request: request
-                )
-            } else {
-                // No refresh token or id token or valid access token, user must reauthenticate
-                NotificationCenter.default.post(name: .reauthenticationRequired)
-                throw RefreshTokenExchangeError.reauthenticationRequired
+            return try await self.serialTaskQueue.enqueue {
```
chionia-gds, line 40: why are we adding the whole check to the queue rather than just the
```
try await self.performRefreshExchangeAndSaveTokens(
                        idToken: tokens.idToken,
                        refreshToken: tokens.refreshToken
                    )
                    
                    return try await self.networkClient.makeAuthorizedRequest(
                        scope: scope,
                        request: request
                    )
```

qnoid: The `if let tokens = try self.sessionManager.validTokensForRefreshExchange` code is the part that reads the token. The read must also be protected.

shukriahmed-dd: Does the sessionManager only read the tokens when its their turn in the queue? So request will join the queue but its not actioned until its turn :) So the refresh exchange will read and overwrite a token then the next refresh exchange will take place

shukriahmed-dd: Nvm the comments in the SerialTaskQueue answer this, thank you !

Tests/UnitTests/Mocks/Sessions+Services/Storage/PersistentSessionManager+Mocks.swift
```diff
@@ -0,0 +1,62 @@
...
+            encryptedStore: mockEncryptedStore,
+            unprotectedStore: mockUnprotectedStore,
+            localAuthentication: mockLocalAuthentication,
+            analyticsService: mockAnalyticsService,
+            walletSDK: mockWalletSDK,
+            serialTaskQueue: serialTaskQueue
+        )
+    }
+    
+    static func resumeSessionPersistentSessionManager(mockAccessControlEncryptedStore: MockSecureStoreService = MockSecureStoreService(),
```
shukriahmed-dd: This function is similar to `setUpNeededForResumeSession` thats available in PersistentManagerTests. And the `make` function above is similar to `setUp` . Looking at test `test_refreshTokenExchange_isSerialisedAcrossResumeSessionAndAuthorizedRequest()` it looks we can use the functions already available in the file 🤔

shukriahmed-dd: I would prefer we keep the same code style we already have the in repo, with setUp being used to create the SUT (of course sometimes for the test we need different properties being passed in - and if its for only 1 test the SUT should be made locally within the test) and functions needed being added within extensions

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
```diff
@@ -888,6 +917,32 @@ extension PersistentSessionManagerTests {
...
+        
+        var keysAsData = String()
+        do {
+            let encoder = JSONEncoder()
+            encoder.outputFormatting = .sortedKeys
+            keysAsData = try encoder.encode(storedTokens).base64EncodedString()
+        } catch {
+            print("error")
+        }
+        return keysAsData
```
shukriahmed-dd: same here, we have `encodeKeys` function below - it would be nice to use these instead and make necessary changes needed

Sources/Utilities/Concurrency/SerialTaskQueue.swift
```diff
@@ -0,0 +1,87 @@
...
+///     for _ in 1...numberOfIncrements {
+///         group.addTask {
+///                 try? await serialTaskQueue.enqueue {
+///                     return await counter.increment()
+///                 }
+///             }
+///         }
+///     }
+/// ```
+///
```
shukriahmed-dd, line 58: This explanation is very helpful, thank you !

Tests/UnitTests/Mocks/Sessions+Services/Storage/PersistentSessionManager+Mocks.swift
qnoid: I would very much like the codebase to [move away](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/792#discussion_r3162775801) from the setUp functions.

Re the `setUpNeededForResumeSession`, I don't think this should be a function in the test case for the reason that it cannot be re-used outside of the context of a test case. This "setup" code is better suited in builder/factory function that create mocks. This will allow mocks to be re-used across test functions and test cases.

The only reason I did not migrate the `setUpNeededForResumeSession` code to the new `make` is because I didn't want to make unnecessary changes for the purposes of this PR.

With that in mind, I do agree it's a good idea to reuse code that builds mocks as much as possible.

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
qnoid: How about I create a debt ticket to make all these changes instead? This way we don't hold this PR back.

Tests/UnitTests/Mocks/Sessions+Services/Storage/PersistentSessionManager+Mocks.swift
shukriahmed-dd: With `setUpNeededForResumeSession` it hasn't been moved because there hasn't been a need to reuse it outside of PersistentManagerTests and I dont imagine it would need to be.

shukriahmed-dd: I think the readability of theses tests is low and I wouldn't be comfortable adding it to the codebase, this PR  creates another MockRefreshToken, I can imagine in future someone would be confuse as to why we needed all 3 and would spend time understanding there different purposes

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
shukriahmed-dd: Since the ticket is low priority I think we dont need to worry about holding it back :)

Sources/Utilities/Concurrency/SerialTaskQueue.swift
```diff
@@ -0,0 +1,87 @@
...
+    private var currentTask: Task<Void, Never>?
+    
+    /// Schedule an `async` operation to start running.
+    ///
+    /// A subsequent call to`enqueue(_:)` will wait starting the operation until the last one to be enqueued has finished executing either succesfuly or with a result value.
+    ///
+    /// - parameter operation: the operation to perfom
+    func enqueue<Success>(_ operation: sending @escaping Operation<Success>) async throws -> Success {
+        try await withCheckedThrowingContinuation { continuation in
+            let newTask = Task { [lastTask = currentTask] in
```
dylanjames-dd, line 73: Can I ask why you've assigned `currentTask` to the `lastTask` local variable in the capture list?

qnoid: This is required to eliminate a deadlock on the actor that would have been created between:

* **A calling thread** on `enqueue`, that has acquired a lock on the actor,  waiting to resume the continuation.
* **A Task thread** accessing the `currentTask` via the actor (i.e. `self`) waiting for the lock to be released.

The capture list removes the need to access the `currentTask` via the actor.

dylanjames-dd: Wow, enlightening 🙂

Tests/UnitTests/Mocks/Sessions+Services/Storage/PersistentSessionManager+Mocks.swift
qnoid: I will add documentation on the `MockRefreshTokenExchangeManagerGuarantor` to explain what it does and how it's meant to be used. It's very purposefully built to trap and report the re-use of a refresh token early. It also very closely emulates what a real network request to the refresh exchange endpoint would feel like by adding a random delay and by adhering strictly to returning a unique refresh token each time.

Tests/UnitTests/Utilities/Session/PersistentSessionManagerTests.swift
qnoid: As a general remark, I don't think the priority of the ticket should affect the priority of getting a PR merged. The goal should be to get PRs merged as soon as possible. The longer it takes for a code change to be merged, the more overheard cost it incurs as `develop` accumulates changes. There is also a the mental and time cost associated with keeping track of a PR, its reviews and changes over time.

shukriahmed-dd (approved):
Looks good! I tested the changes in the build app and expired the access token so refresh token exchange can take place :)

qnoid:
@shukriahmed-dd thank you Shukri for testing in the app! Greatly appreciated.

chionia-gds (approved):

PR #812, opened 2026-05-14
feat: DCMAW-20361 update wallet version to 16.139.0 and use the new config

chionia-gds:
# DCMAW-20361 update wallet version to 16.139.0 and use the new config

## If wallet store ID or Wallet environment are nil during initialisation:

https://github.com/user-attachments/assets/f9cebcdd-a5d2-4006-a0a3-00334b5ee9b8

Above will be updated in https://govukverify.atlassian.net/browse/DCMAW-20468

# Checklist

### DIFF
--- a/Sources/Tabs/WalletCoordinator.swift
+++ b/Sources/Tabs/WalletCoordinator.swift
@@ -46,7 +46,14 @@ final class WalletCoordinator: NSObject,
         
         guard let walletStoreID = sessionManager.walletStoreID,
               let walletEnvironment = WalletEnvironment(buildConfiguration: AppEnvironment.buildConfiguration.lowercased()) else {
-            fatalError("walletStoreID or walletEnvironment was not initialised and was nil")
+            // TODO: DCMAW-20468 update this with new error screen with relevant text
+            let viewModel = UnrecoverableLoginErrorViewModel(
+                analyticsService: analyticsService,
+                errorDescription: "walletStoreID or walletEnvironment was not initialised and was nil"
+            )
+            let walletNotInitialisedErrorScreen = GDSErrorScreen(viewModel: viewModel)
+            root.pushViewController(walletNotInitialisedErrorScreen, animated: false)
+            return
         }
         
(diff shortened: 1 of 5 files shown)


shukriahmed-dd (approved):

Sources/Tabs/WalletCoordinator.swift
```diff
@@ -43,10 +43,16 @@ final class WalletCoordinator: NSObject,
             image: UIImage(systemName: "wallet.pass"),
             tag: 1
         )
-        let walletConfig = WalletConfigV2(
-            environment: WalletEnvironment(buildConfiguration: AppEnvironment.buildConfiguration.lowercased()),
+        
+        guard let walletStoreID = sessionManager.walletStoreID,
+              let walletEnvironment = WalletEnvironment(buildConfiguration: AppEnvironment.buildConfiguration.lowercased()) else {
+            fatalError("walletStoreID or walletEnvironment was not initialised and was nil")
```
qnoid: I believe this will crash the app. Do we want that in production? I think it would be preferable to disable wallet and log a "crash" to analytics so that we can investigate?

chionia-gds: Dylan and Jamie confirmed that if either of those are ever missing we need to crash the app because they should never be missing.

qnoid: I don't think crashing the app, in production, is a good idea. An app that terminates unexpectedly can lead to data loss, data corruption and it is a bad use experience.

Is there a way to support graceful degradation instead? Effectively the app is able to continue to run but the user cannot add any credentials?

chionia-gds: okay I had a chat with Vit and told him to make a new product requirement for this: https://govukverify.atlassian.net/browse/DCMAW-20468
We will continue with this existing error screen but will update after this ticket gets approved by UCD

qnoid (approved):

PR #833, opened 2026-06-09
feat: DCMAW-20314 fix a case of user stuck in a login loop

qnoid:
# [iOS | Users stuck in a login loop ](https://govukverify.atlassian.net/browse/DCMAW-20314)

This PR fixes a case where a user is stuck in a loop between two screens:

“You need to sign in again”
“There was a problem signing you in”

by clearing the session data for the user even when wallet throws an error.  Without this change, **the user has no choice but to delete the app**.

# Discussion

This PR proposes that in the case where `WalletSessionData` throws `.cannotDeleteData` , the rest of the `SessionBoundData` proceeds with deletion of the rest of the user data, including the `UserDefaults` which, **as far as I can tell**, is actually the backing store for the [SecureStoreService](https://github.com/govuk-one-login/mobile-ios-secure-store/blob/3.0.0/Sources/SecureStore/SecureStoreService.swift#L13) which throws the error when failing to decrypt a stored entry. For a more in depth explanation, see [DCMAW-20314: Technical Report: Users stuck in a login loop](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/6652264520/DCMAW-20314+Technical+Report+Users+stuck+in+a+login+loop)

# Changes

## `WebAuthenticationService` error handling

This required a **change in the current implementation** of the `WebAuthenticationService`; more specifically, in case a `PersistentSessionError` of  `.cannotDeleteData` case is thrown, the `WebAuthenticationService` **will clear all session data only this time opting not to fail early  in case of an error**, like `startAuthSession` does for a returning  user that is NOT authenticated.

The intent of this change is to recover from an invalid state by  primarily clearing the data in the `UserDefaultsStore`, which is what the `SecureStoreService` uses as its backing store, which throws a `WalletError` of `.failedToDeleteProofKeys` case due to failing to decrypt the value, as it was read by the `UserDefaultsStore`, in the case of a backup that has been restored which resulted in the encryption key to go missing.

The behaviour is otherwise unchanged, including logging and throwing the error as it was previously handled by the `default` case.

## `UserDefaults` and `clearAllSessionData(presentSystemLogOut:)`
In order to be able to test that `clearSessionData` is [called on](https://github.com/govuk-one-login/mobile-ios-one-login-app/compare/fix/DCMAW-20314-user-stuck-in-a-login-loop?expand=1#diff-56138c3a6ba90730645170a31a2030a44e239fe859f259aa96a911066ae40896R250) the `UserDefaults`, changed `registerSessionBoundData` to clear the `unprotectedStore` in place of the user defaults given that `unprotectedStore` is a reference to `UserDefaults.standard`.

# Where to focus the review

* Should the scope of the catch go further to `WalletError` of `.failedToDeleteProofKeys` kind or should it be kept at a high level to the `PersistentSessionError` of `.cannotDeleteData` kind? The former will make the change surgically more precise without risking any unattended side effects. The latter broadens the scope so that the code can recover from any reason a `PersistentSessionError` is thrown for `.cannotDeleteData`. As a side note, as far as I can see, limiting the scope to `WalletError` of `.failedToDeleteProofKeys`, will require bringing in `Storage` as a dependency where the `WalletStore` (i.e. `.failedToDeleteProofKeys`) enum is defined.

* Even though the proposed code change will remedy the situation, by removing all user session data, is there a chance that the app will be left in a different state that is also invalid? **Should** removing all user session data instead **be done in reverse**? i.e. where `WalletSessionData` is deleted last, thus more likely to succeed given that the storage (i.e. `UserDefaults`) of `SecureStoreService` has now been cleared? 

* Even tho this change alters the behaviour of the app, only for the case of `.cannotDeleteData`, it still represents a significant change compared to how the app already behaves. That is, "remove all user session data (including the persistent session and Wallet data) from the device" but do so "atomically" where wallet data is the first one to be attempted. Should that fail, do not proceed with the deletion. What risks does that carry?

# Testing

Once this change has been deployed to staging and a build is available on TestFlight, will need to follow the "Steps to reproduce" as described in the ticket to confirm the fix.

Do you have any suggestions of either other unit tests or manual testing that should be done as part of running a regression test suite?

# Related Reading

[DCMAW-20314: Technical Report: Users stuck in a login loop](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/6652264520/DCMAW-20314+Technical+Report+Users+stuck+in+a+login+loop)

# Checklist

### DIFF
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -7,6 +7,7 @@ import GDSUtilities
 import LocalAuthenticationWrapper
 import Logging
 import SecureStore
+import WalletStore
 
 // swiftlint:disable:next type_body_length
 final class PersistentSessionManager: SessionManager {
@@ -202,6 +203,18 @@ final class PersistentSessionManager: SessionManager {
         (try? localAuthentication.canUseAnyLocalAuth) ?? false && isReturningUser
     }
     
+    private func clear(sessionData: [SessionBoundData], presentSystemLogOut: Bool) async throws {
+        for each in sessionData {
+            try await each.clearSessionData()
+        }
+        
+        endCurrentSession()
+        
+        if presentSystemLogOut {
+            NotificationCenter.default.post(name: .systemLogUserOut)
+        }
+    }
+
     func startAuthSession(
         _ session: any LoginSession,
         using configuration: @Sendable (String?) async throws -> LoginSessionConfiguration
@@ -218,6 +231,13 @@ final class PersistentSessionManager: SessionManager {
                                                                          reason: "secure wallet data deleted"))
                     }
                     try await clearAllSessionData(presentSystemLogOut: true)
+                } catch let error as WalletStoreError where error.kind == .failedToDeleteProofKeys {
+                    do {
+                        try await clear(sessionData: self.sessionBoundData.reversed(), presentSystemLogOut: true)
+                    } catch {
+                        analyticsService.logCrash(error)
+                    }
+                    throw PersistentSessionError(.cannotDeleteData, originalError: error)
                 } catch {
                     throw PersistentSessionError(.cannotDeleteData, originalError: error)
                 }
@@ -390,15 +410,7 @@ final class PersistentSessionManager: SessionManager {
     }
 
     func clearAllSessionData(presentSystemLogOut: Bool) async throws {
-        for each in sessionBoundData {
-            try await each.clearSessionData()
-        }
-        
-        endCurrentSession()
-        
-        if presentSystemLogOut {
-            NotificationCenter.default.post(name: .systemLogUserOut)
-        }
+        try await self.clear(sessionData: sessionBoundData, presentSystemLogOut: presentSystemLogOut)
     }
     
     func registerSessionBoundData(_ data: [SessionBoundData]) {
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -203,18 +203,6 @@ final class PersistentSessionManager: SessionManager {
         (try? localAuthentication.canUseAnyLocalAuth) ?? false && isReturningUser
     }
     
-    private func clear(sessionData: [SessionBoundData], presentSystemLogOut: Bool) async throws {
-        for each in sessionData {
-            try await each.clearSessionData()
-        }
-        
-        endCurrentSession()
-        
-        if presentSystemLogOut {
-            NotificationCenter.default.post(name: .systemLogUserOut)
-        }
-    }
-
     func startAuthSession(
         _ session: any LoginSession,
         using configuration: @Sendable (String?) async throws -> LoginSessionConfiguration
@@ -233,7 +221,7 @@ final class PersistentSessionManager: SessionManager {
                     try await clearAllSessionData(presentSystemLogOut: true)
                 } catch let error as WalletStoreError where error.kind == .failedToDeleteProofKeys {
                     do {
-                        try await clear(sessionData: self.sessionBoundData.reversed(), presentSystemLogOut: true)
+                        try await clearSessionData(in: self.sessionBoundData.reversed(), presentSystemLogOut: true)
                     } catch {
                         analyticsService.logCrash(error)
                     }
@@ -402,17 +390,26 @@ final class PersistentSessionManager: SessionManager {
     }
     
     func clearAppForLogin() async throws {
-        for each in sessionBoundData where type(of: each) != UserDefaultsPreferenceStore.self {
+        let excludingUserDefaultsPreferenceStore = sessionBoundData.filter { type(of: $0 ) != UserDefaultsPreferenceStore.self }
+        try await self.clearSessionData(in: excludingUserDefaultsPreferenceStore, presentSystemLogOut: false)
+    }
+
+    func clearAllSessionData(presentSystemLogOut: Bool) async throws {
+        try await self.clearSessionData(in: sessionBoundData, presentSystemLogOut: presentSystemLogOut)
+    }
+    
+    private func clearSessionData(in sessionData: [SessionBoundData], presentSystemLogOut: Bool) async throws {
+        for each in sessionData {
             try await each.clearSessionData()
         }
         
         endCurrentSession()
+        
+        if presentSystemLogOut {
+            NotificationCenter.default.post(name: .systemLogUserOut)
+        }
     }
 
-    func clearAllSessionData(presentSystemLogOut: Bool) async throws {
-        try await self.clear(sessionData: sessionBoundData, presentSystemLogOut: presentSystemLogOut)
-    }
-    
     func registerSessionBoundData(_ data: [SessionBoundData]) {
         sessionBoundData = data
     }
--- a/Tests/UnitTests/Mocks/Sessions+Services/Storage/MockDefaultsStore.swift
+++ b/Tests/UnitTests/Mocks/Sessions+Services/Storage/MockDefaultsStore.swift
@@ -24,3 +24,36 @@ class MockDefaultsStore: DefaultsStoring, SessionBoundData {
         savedData = [:]
     }
 }
+
+struct MockDefaultsStoreExpectation: DefaultsStoring, SessionBoundData {
+    typealias ClearSessionDataAsFunction = () throws -> Void
+    
+    let mockDefaultsStore: MockDefaultsStore
+    let clearSessionDataAsFunction: ClearSessionDataAsFunction
+
+    init(mockDefaultsStore: MockDefaultsStore = MockDefaultsStore(), clearSessionDataAsFunction: @escaping ClearSessionDataAsFunction = { }) {
+        self.mockDefaultsStore = mockDefaultsStore
+        self.clearSessionDataAsFunction = clearSessionDataAsFunction
+    }
+    
+    func set(_ value: Any?, forKey defaultName: String) {
+        self.mockDefaultsStore.set(value, forKey: defaultName)
+    }
+    
+    func value(forKey key: String) -> Any? {
+        return self.mockDefaultsStore.value(forKey: key)
+    }
+    
+    func bool(forKey: String) -> Bool {
+        return self.mockDefaultsStore.bool(forKey: forKey)
+    }
+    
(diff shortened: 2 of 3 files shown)


Sources/Login/WebAuthenticationService.swift
```diff
@@ -49,6 +49,10 @@ final class WebAuthenticationService: AuthenticationService {
                 analyticsService.logCrash(error)
             }
             throw error
+        } catch let error as PersistentSessionError where error.kind == .cannotDeleteData {
+            await sessionManager.clearAllSessionDataIgnoreErrors(presentSystemLogOut: true)
```
chionia-gds: my only concern is that this could delete things on our side and then something from wallet could fail to delete and then a new user would still have wallet keys from the previous user in the secure store. Like the proof keys error we were getting from. wallet

qnoid: Is there a specific scenario you can think of?

Alternatively, we could make the following changes:

* Further limit the scope of the `clearAllSessionDataIgnoreErrors` to only a `WalletError` of `.failedToDeleteProofKeys` key like so:

```
catch let error as PersistentSessionError where error.kind == .cannotDeleteData {
            if case let walletError = error.originalError as? WalletError<WalletStore>, walletError.kind == .failedToDeleteProofKeys  {
                await sessionManager.clearAllSessionDataIgnoreErrors(presentSystemLogOut: true)
            }
            
            analyticsService.logCrash(error)
            throw error
        } 
```

* `clearAllSessionDataIgnoreErrors` to reverse the session data, so they are deleted in the following order:

1.  user defaults first (which is where the data is located that `SecureStoreService` tries to delete and throws the error.
2. ...
3. ...
4. wallet session data, which will now succeed, given that the user defaults has been deleted

qnoid: We could narrow the scope even further by moving the catch inside the `PersistentSessionManager` for this specific code path (returning user, not authenticated). Effectively, [at the point the code catches](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.24.0/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L182) the `WalletError`, before throwing the `PersistentSessionError`.

Tests/UnitTests/Mocks/Sessions+Services/Storage/MockDefaultsStore.swift
```diff
@@ -24,3 +24,37 @@ class MockDefaultsStore: DefaultsStoring, SessionBoundData {
...
+    
+    func removeObject(forKey defaultName: String) {
+        self.mockDefaultsStore.removeObject(forKey: defaultName)
+    }
+    
+    func clearSessionData() throws {
+        try self.mockDefaultsStore.clearSessionData()
+        try self.clearSessionDataAsFunction()
+    }
+
```
shukriahmed-dd: nit extra line

qnoid:
I have separated the changes between 2 commits:
* Make `WebAuthenticationService` and `PersistentSessionManager` testable (these changes have since been merged to develop and not part of this PR)
* [The proposed fix for the defect.](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833/changes/3f58117682c375b64b1900b3f758bfdc5cc3a4aa)

The former changes are now tracked by a separate PR at https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/834

Sources/Login/WebAuthenticationService.swift
robinphillips: The only reason Wallet exposes these errors to the app is for logging purposes. If Wallet had an analytics service at the point of these delete calls being made, Wallet would log them all internally and not expose the error to One Login.

Sources/Login/WebAuthenticationService.swift
```diff
@@ -49,6 +58,10 @@ final class WebAuthenticationService: AuthenticationService {
                 analyticsService.logCrash(error)
             }
             throw error
+        } catch let error as PersistentSessionError where error.kind == .cannotDeleteData {
```
robinphillips: Are all the other changes in the PR required for resolving this defect? As I understood it, the reason users were stuck in a loop was entirely because of handling Wallet errors and showing an error screen to users.

I'm not so familiar with the One Login codebase, but it seems to me that this is the place that needs fixing and much of the rest of the changes in this PR is unrelated to the specific defect.

qnoid: You can see the targeted changes to fix this defect in this [commit](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833/changes/b1933ece0202570fc57d8fd4b7c9dece29db1be9).

Was there a specific one you have in mind? The other changes are primarily required to be able to write a test for this fix. There shouldn't be any other changes.

Sources/Login/WebAuthenticationService.swift
qnoid: I understand. In that case, it would not be possible to suggest a fix within One Login for this user scenario. In either case, the user is blocked from making forward progress.

qnoid:
Have now updated the PR following discussion, feedback and more input.

Here are the answers to the questions posed in this PR.

> **Q: Should the scope of the catch go further to WalletError of .failedToDeleteProofKeys kind or should it be kept at a high level to the PersistentSessionError of .cannotDeleteData kind?** 
> A: In order to prevent any side effects this fix will be targeted, narrow and tightly coupled to the Wallet implementation. This trades the risk for regressions caused by this change **today** vs a future regression by changes in the Wallet implementation. 

> **Q: Even though the proposed code change will remedy the situation, by removing all user session data, is there a chance that the app will be left in a different state that is also invalid? Should removing all user session data instead be done in reverse? i.e. where WalletSessionData is deleted last, thus more likely to succeed given that the storage (i.e. UserDefaults) of SecureStoreService has now been cleared?**
> A: The [user session data will now be cleared in reverse](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833/changes). This assumes that by clearing user defaults, Wallet will not throw an error when attempted to clear. This is deemed lower risk than silencing the error that Wallet throws since no data will be left behind. The precedent of One Login clearing Wallet data is already set. Furthermore, in this specific case of a returning user who is NOT authenticated, the following are true:
    1. The current implementation code already expects the data to be deleted
    2. Leaving wallet data behind carries higher risk than not deleting it
    3. The PR assumes the delete code to succeed when the order is in reverse. 
    4. There is a risk that wallet is left in an invalid state if we don’t attempt to clear


> **Q: Even tho this change alters the behaviour of the app, only for the case of .cannotDeleteData, it still represents a significant change compared to how the app already behaves. That is, "remove all user session data (including the persistent session and Wallet data) from the device" but do so "atomically" where wallet data is the first one to be attempted. Should that fail, do not proceed with the deletion. What risks does that carry?**
> A: The data will be deleted in reverse order on the basis that by clearing user data first, wallet will not throw an error.

qnoid:
Now that https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/834 has been merged to develop, this PR only holds the [changes](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833/changes) to fix the defect.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -202,6 +203,18 @@ final class PersistentSessionManager: SessionManager {
         (try? localAuthentication.canUseAnyLocalAuth) ?? false && isReturningUser
     }
     
+    private func clear(sessionData: [SessionBoundData], presentSystemLogOut: Bool) async throws {
```
chionia-gds: should we move this function below the other clearing ones? And maybe name it something more specific like "clearSessionData(in sessionData: [SessionBoundData], ...)"

qnoid: I was under the impression that private functions go to the top? Happy to move it closer to the other ones if that's the exception to the rule. 

Will adopt the name suggestion too.

chionia-gds: We don't follow that layout I don't think, if anything in most files in the repo they are at the bottom

qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833/changes/2bb7b18f478e9eabc3ec91567f7364691055b710

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -218,6 +219,13 @@ final class PersistentSessionManager: SessionManager {
                                                                          reason: "secure wallet data deleted"))
                     }
                     try await clearAllSessionData(presentSystemLogOut: true)
+                } catch let error as WalletStoreError where error.kind == .failedToDeleteProofKeys {
+                    do {
+                        try await clearSessionData(in: self.sessionBoundData.reversed(), presentSystemLogOut: true)
```
shukriahmed-dd, line 224: why is the data reversed?

qnoid: @shukriahmed-dd [see this comment](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833#issuecomment-4689623292). If you want to know how/why that is expected to work, the [technical report](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/6652264520/DCMAW-20314+Technical+Report+Users+stuck+in+a+login+loop) goes into more detail.

shukriahmed-dd: thank you ! that makes sense, I wonder if it worth adding a comment in the code e.g. `// Data's order is reserved, so UserDefaults is cleared first to prevent Wallet from throwing an error whilst clearing `

shukriahmed-dd (approved):
looks good

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: I don't think that comment is very useful in isolation without the context surrounding it. Once the public contract has been formalised it will be useful to have documentation as to what the code does.

As a general remark, I use `git blame` to understand the context of why a change was made. This is another reason to have good git comments and a PR that includes a healthy discussion with decisions made on the record. That way someone can look back and understand why and under what circumstances decisions were made; it is also easier to make an informed decision and make changes should the circumstances have changed since.

jamielutton-gds (approved):

qnoid:
🚀

PR #875, opened 2026-07-16
feat: DCMAW-21030 fix a case of a user stuck in a login loop

qnoid:
# [iOS | Tech Debt | SecureStoreError(.cantDecryptData) | NSOStatusErrorDomain (-50) | KeyManagerService](https://govukverify.atlassian.net/browse/DCMAW-21030)

> [!IMPORTANT]
> There are no steps to reproduce this defect. This change has not been tested on a device.

This PR fixes a case where a user is stuck in a loop between two screens:

“You need to sign in again”
“Sorry, there's a problem”

when the app attempts to add an attestation header as part of a token exchange. Without this change, the user has no choice but to delete the app.

test_signInWarningScreen 1 test_genericError 1

# Preface

Here is the stack trace as identified after a code review:

- [LoginCoordinator.start()](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/LoginCoordinator.swift#L67)
- [LoginCoordinator.authenticate](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/LoginCoordinator.swift#L91)
- [LoginCoordinator.launchAuthenticationService](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/LoginCoordinator.swift#L98C27-L98C42)
- [LoginCoordinator.triggerAuthFlow](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/LoginCoordinator.swift#L116)
- - [WebAuthenticationService.startWebSession()](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/WebAuthenticationService.swift#L26) // [FirebaseAppIntegrityService](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Extensions/CustomTypes/AppIntegrityProvider%2BOneLogin.swift#L24) used to provide attestation header is created here. The attestation is stored in the [SecureAttestationStore](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Utilities/Storage/SecureAttestationStore.swift)
- - - [PersistentSessionManager.startAuthSession()](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L243) // In this case, `persistentID` is nil because it can't be decrypted, wallet data is cleared without throwing an error)
- - - -[AppAuthSession.performLoginFlow()](https://github.com/govuk-one-login/mobile-ios-authentication/blob/main/Sources/Authentication/AppAuthSession.swift#L59) // user logs in succesfully
- - - - [AppAuthSession.performLoginFlow()](https://github.com/govuk-one-login/mobile-ios-authentication/blob/main/Sources/Authentication/AppAuthSession.swift#L65C27-L65C56) // Token exchange is attempted
- - - - [AppAuthSession.finaliseLoginWithAuthResponse()](https://github.com/govuk-one-login/mobile-ios-authentication/blob/main/Sources/Authentication/AppAuthSession.swift#L115)
- - - - [AppAuthSession.handleAuthResponseCreateTokenRequest()](https://github.com/govuk-one-login/mobile-ios-authentication/blob/main/Sources/Authentication/AppAuthSession.swift#L152) // attempt to evaluate token headers
- - - - - [LoginSessionConfiguration+OneLogin.oneLoginSessionConfiguration()](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Extensions/CustomTypes/LoginSessionConfiguration%2BOneLogin.swift#L25) // Attempt to get token headers from integrityAssertions
- - - - - [FirebaseAppIntegrityService.integrityAssertions](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift#L241) // Attempts to read attestation in case it has either expired or has never been fetched before.
- - - - - - [SecureAttestationStore](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Utilities/Storage/SecureAttestationStore.swift#L27)
- - - - - - - [SecureStoreService.readItem](https://github.com/govuk-one-login/mobile-ios-secure-store/blob/4.0.0/Sources/SecureStore/SecureStoreService.swift#L35) // Attempts to decrypt attestation
- - - - - - - - [KeyManagerService.decryptDataWithPrivateKey()](https://github.com/govuk-one-login/mobile-ios-secure-store/blob/main/Sources/SecureStore/KeyManagerService.swift#L197) // The new key pair that was created for the instance of the KeyManagerService is not the same as the one used to encrypt the data before the backup.
- - - - - - - - [KeyManagerService.biometricErrorHandling()](https://github.com/govuk-one-login/mobile-ios-secure-store/blob/main/Sources/SecureStore/SecureStoreError/SecureStoreError.swift#L58) // throws ``SecureStoreError(
                    .cantDecryptData)`` with original error -50.
...
- [LoginCoordinator.catch](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/LoginCoordinator.swift#L108)
- [LoginCoordinator.showGenericErrorScreen](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Login/LoginCoordinator.swift#L300)

# Changes

> [!TIP]
> This PR includes 2 commits. 
> 1. [The first one](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/875/changes/5b7f6047b28e50cf9afafa83f6ae28f6b59fb13f), does not modify the source code in any way that it alters its behaviour. It merely adds a new function, [SecureAttestationStore.make()](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/875/changes/5b7f6047b28e50cf9afafa83f6ae28f6b59fb13f#diff-6c2d8f430ec91670e705f386ba5a828c2bd1aee1d165190addc6ca6a40b50ce9R29), that the second commit uses to fix the defect. Other than that, the commit includes Tests, Mocks and makes minor changes (e.g. elevate access to the [AppCheckVendor](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/875/changes/5b7f6047b28e50cf9afafa83f6ae28f6b59fb13f#diff-5706497a63689b6917dcb7e7f4056829d83d5069ea8974e176efa58828f994e4R159) allow testing of the [FirebaseAppIntegrityService.swift](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/875/changes/5b7f6047b28e50cf9afafa83f6ae28f6b59fb13f#diff-5706497a63689b6917dcb7e7f4056829d83d5069ea8974e176efa58828f994e4R159).
> 2. [The second commit](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/875/changes/528896028e02e243a3b8e16d21c7ef7ba9d84b0d), uses the `SecureAttestationStore.make()` to fix the defect.

This PR effectively [makes a single](https://github.com/govuk-one-login/mobile-ios-one-login-app/commit/528896028e02e243a3b8e16d21c7ef7ba9d84b0d) but crucial code change in the source code that fixes a case of a user stuck in a login loop. More specifically, in case the underlying **attestation data** cannot be decrypted deletes it.

# Discussion

> [!IMPORTANT]
> I don't believe any users are currently affected by this defect. That is because, when the `persistentId` is evaluated to `nil` (due to the -50 underlying error) the `PersistentSessionManager` attempts to clear the Wallet session data, which in turns throws an error. This is a [known defect ](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833). I expect once WalletSDK has fixed the issue, we will encounter this defect and the occurrences will increase up to the number of users that have restored their device from a backup.

This PR makes the following proposals:
1. Detect a case of cantDecryptData as early as possible (i.e. when the `SecureAttestationStore` is created) so as to avoid any unexpected behaviour in the app (as in this case)
2. In the case the data cannot be decrypted, delete it (attestation and expiry date) associated with the `SecureAttestationStore` so as the app works as expected.

# Where to focus the review
There are a couple of things that are hard to test within the codebase and would benefit from the code review.
* The `FirebaseAppIntegrityService` instance created by the `AppIntegrityProvider.firebaseAppCheck()` cannot be tested due to the fact that the `AppCheckVendor` instance is evaluated to `AppCheck.appCheck()`. That would require importing `Firebase` et. al. directly to OneLogin. **I am not sure we want that direct dependency**. As a result, the `FirebaseAppIntegrityService` instance that is provided by the `FirebaseAppIntegrityService+Mocks.swift` is **not the same** as the one used by the codebase. Having said that, the **only change**  in the `FirebaseAppIntegrityService` that needs to be made is to ensure that it it uses the `attestationStore` as returned by `SecureAttestationStore.make()`. We should aim to make the `AppIntegrityProvider.firebaseAppCheck()` testable so as to prevent any future regressions.
* Test coverage has been added to the instance returned by `SecureAttestationStore.make()`. However, given how the instance returned changes the post-conditions (i.e. both the attestationJWT and expiry date are removed) and the high coupling of the `SecureAttestationStore` (e.g. used to create token headers), it's worth reviewing the code to ensure the invariants of the `SecureAttestationStore` are preserved. e.g. notice how the `SecureStoreService` throws a `SecureStoreError(.unableToRetrieveFromUserDefaults)` in case either the `attestationJWT` or the `attestationExpired` is missing.

Looking at the `FirebaseAppIntegrityService/clientAssertions`. The `FirebaseAppIntegrityService` guards on `hasExpiredAttestation`. The `hasExpiredAttestation` property is read from `SecureAttestationStore/attestationExpired`. In case the expiry date is missing, `FirebaseAppIntegrityService` will treat the attestation as expired. 

As far as I can see:
* In case the `attestationExpired` is **missing**, the error is swallowed and the `attestationJWT` is considered to have expired.
* In this case, the attestation has expired  and the `attestationJWT` will fetched again.

# Testing

> [!IMPORTANT]
> As of this moment, there are no steps to reproduce this within the app. Instead, this PR is the product of a code review that lead to two test cases written that hint at how it is possible for a user to get stuck in this loop.

The following tests have been added:

* test_errorFromAttestationJWT_onNewUser
* test_errorFromAttestationJWT_onReturningUser

As well as:

* test_attestionJWT_throwsCantDecryptDataError
* test_attestionJWT_doesNotThrowCantDecryptdataError
 
The `test_attestionJWT_doesNotThrowCantDecryptdataError` test was written against the instance of the `SecureAttestationStore` as returned by its initialiser, which caused it to fail. Changing the instance to the one returned by the `SecureAttestationStore.make()`, introduced in this PR, made the test pass.

DCMAW-21030_failure

DCMAW-21030_success

# Checklist

### DIFF
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -27,8 +27,10 @@ final class PersistentSessionManager: SessionManager {
         
         let encryptedSecureStoreMigrator = encryptedStore ?? EncryptedSecureStoreMigrator(analyticsService: analyticsService)
         let manager = PersistentSessionManager(
-            accessControlEncryptedStore: accessControlEncryptedSecureStoreMigrator,
             encryptedStore: encryptedSecureStoreMigrator,
+            storeKeyService: SecureTokenStore(
+                accessControlEncryptedStore: accessControlEncryptedSecureStoreMigrator
+            ),
             unprotectedStore: unprotectedStore,
             localAuthentication: LocalAuthenticationWrapper(localAuthStrings: .oneLogin),
             analyticsService: analyticsService,
@@ -50,7 +52,6 @@ final class PersistentSessionManager: SessionManager {
         return manager
     }
     
-    private let accessControlEncryptedStore: SecureStorable
     private let encryptedStore: SecureStorable
     private let storeKeyService: TokenStore
     private let unprotectedStore: DefaultsStoring
@@ -70,15 +71,15 @@ final class PersistentSessionManager: SessionManager {
     let serialTaskQueue: SerialTaskQueue
     
     convenience init(
-        accessControlEncryptedStore: SecureStorable,
         encryptedStore: SecureStorable,
+        storeKeyService: SecureTokenStore,
         analyticsService: OneLoginAnalyticsService,
         tokenExchangeManager: TokenExchangeManaging,
         serialTaskQueue: SerialTaskQueue = SerialTaskQueue(),
     ) {
         self.init(
-            accessControlEncryptedStore: accessControlEncryptedStore,
             encryptedStore: encryptedStore,
+            storeKeyService: storeKeyService,
             unprotectedStore: UserDefaults.standard,
             localAuthentication: LocalAuthenticationWrapper(localAuthStrings: .oneLogin),
             analyticsService: analyticsService,
@@ -88,8 +89,8 @@ final class PersistentSessionManager: SessionManager {
     }
     
     init(
-        accessControlEncryptedStore: SecureStorable,
         encryptedStore: SecureStorable,
+        storeKeyService: SecureTokenStore,
         unprotectedStore: DefaultsStoring,
         localAuthentication: LocalAuthManaging,
         analyticsService: OneLoginAnalyticsService,
@@ -97,11 +98,8 @@ final class PersistentSessionManager: SessionManager {
         tokenExchangeManager: TokenExchangeManaging,
         serialTaskQueue: SerialTaskQueue = SerialTaskQueue()
     ) {
-        self.accessControlEncryptedStore = accessControlEncryptedStore
         self.encryptedStore = encryptedStore
-        self.storeKeyService = SecureTokenStore(
-            accessControlEncryptedStore: accessControlEncryptedStore
-        )
+        self.storeKeyService = storeKeyService
         self.unprotectedStore = unprotectedStore
         self.localAuthentication = localAuthentication
         
--- a/AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift
+++ b/AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift
@@ -156,7 +156,7 @@ public final class FirebaseAppIntegrityService: AppIntegrityProvider {
         }
     }
     
-    init(
+    public init(
         vendor: AppCheckVendor,
         attestationProofOfPossessionProvider: ProofOfPossessionProvider,
         attestationProofOfPossessionTokenGenerator: ProofOfPossessionTokenGenerator,
--- a/AppIntegrity/Sources/AppIntegrity/Protocols/AppCheckVendor.swift
+++ b/AppIntegrity/Sources/AppIntegrity/Protocols/AppCheckVendor.swift
@@ -1,6 +1,6 @@
 import FirebaseAppCheck
 
-protocol AppCheckVendor {
+public protocol AppCheckVendor {
     static func setAppCheckProviderFactory(_ factory: AppCheckProviderFactory?)
     static func appCheck() -> Self
 
--- a/OneLogin.xcodeproj/project.pbxproj
+++ b/OneLogin.xcodeproj/project.pbxproj
@@ -256,6 +256,7 @@
 		8E9587DA2F45E6E80073A475 /* AppIntegrityErrorViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E9587D92F45E6E80073A475 /* AppIntegrityErrorViewModelTests.swift */; };
 		8E9A51852FDAEA4500F4C93A /* LoginSessionConfiguration+OneLoginTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E9A51842FDAEA4500F4C93A /* LoginSessionConfiguration+OneLoginTests.swift */; };
 		8EA9FDBB2F3CE17E00C52D4D /* AppIntegrityErrorViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8EA9FDBA2F3CE17E00C52D4D /* AppIntegrityErrorViewModel.swift */; };
+		9996E02F30064DF900AB019B /* FirebaseAppIntegrityService+Mocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9996E02E30064DF400AB019B /* FirebaseAppIntegrityService+Mocks.swift */; };
 		99AEB1C52FFCFB2E00F834BB /* PersistentSessionManager+Mocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 99AEB1C42FFCFB2E00F834BB /* PersistentSessionManager+Mocks.swift */; };
 		AB24C1132DB94A5F0080141C /* LocalAuthSettingsErrorViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = AB24C1122DB948520080141C /* LocalAuthSettingsErrorViewModel.swift */; };
 		AB327B112D36BEAD00AC169E /* CRIOrchestrator in Frameworks */ = {isa = PBXBuildFile; productRef = AB327B102D36BEAD00AC169E /* CRIOrchestrator */; };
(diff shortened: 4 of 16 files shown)


qnoid:
Upon further review of the code, I don't think the code base will be affected by this defect. 
That is because, the `attestationExpired` property in the SecureAttestationStore, [in case of an error it swallows the error and considers the attestation to have expired](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/Sources/Utilities/Storage/SecureAttestationStore.swift#L18). In [which case](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift#L48), the attestation will be [fetched anew](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift#L57). 

In case it is successful, both the expiry date and attestation will [be stored](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.26.0/AppIntegrity/Sources/AppIntegrity/FirebaseAppIntegrityService.swift#L212) and **replace** the existing data (which up until this has been unrecoverable), encrypted with the new key pair. Subsequent calls to read the expiry date and attestation will succeed since the data can now be decrypted.

Having said all that, I still think it's a good idea to accept the fix. For example, the code makes an implicit assumption that the attestationJWT is only valid to be **accessed**  when it has not expired. It also treats an error as an "expired" attestation. I would argue this makes the code fragile and prone for a new defect to be introduced and/or a regression in case:
* The attestationJWT is accessed without first checking the expiry date.
* The attestation is no longer considered "expired" in case of an error

Even though in this case the code is able to recover by replacing the existing inaccessible data, I don't think that case should not handled by the `FirebaseAppIntegrityService`. I think it's best this error is handled and mitigated as early as detected which is at the time the `SecureAttestationStorage` is created. 

PS. I see this as a temporary solution until the work is done to detect a case of inaccessible data at the level of `KeyManagerService`.
PSS. I think the implementation of the `SecureAttestationStore` should make it safer to use and not leave any room for misuse. I see both the `expiryDate` and `attestationJWT` as defining the invariant of the `SecureAttestationStore`. That is to say, either the attestation has expired, in which case there is no attestation or the attestation is still valid and this is the value. In practice that would mean using and exposing an enum to consumers like so:

```swift 
enum Attestation {
case expired
case valid(attestationJWT: String)
}
```

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -27,8 +27,10 @@ final class PersistentSessionManager: SessionManager {
         
         let encryptedSecureStoreMigrator = encryptedStore ?? EncryptedSecureStoreMigrator(analyticsService: analyticsService)
         let manager = PersistentSessionManager(
-            accessControlEncryptedStore: accessControlEncryptedSecureStoreMigrator,
```
shukriahmed-dd, line 30: Is there a benefit to not passing in `accessControlEncryptedStore` or was this not needed ?

qnoid: Well, it is not needed thus there is a benefit in not having to have an instance of the type to create a `PersistentSessionManager`.

shukriahmed-dd (approved):
looks good !

chionia-gds (approved):

PR #890, opened 2026-07-31
feat: DCMAW-21048 ensure user session is evaluated once before attempting to evaluate it a second time

qnoid:
# [Ensure user session is evaluated once before attempting to evaluate it a second time](https://govukverify.atlassian.net/browse/DCMAW-21048)

The proposed code changes ensure that user session is evaluated once before attempting to evaluate it a second time. In practice this code change guarantees that:
1. ~~Evaluating a user session is done **serially**~~ (This work was done as part of https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/892 which this PR is based on)
2. The number of requests to resume a session are **throttled**

# Discussion
To better demonstrate the issue as described in the ticket, there are a number of commits that lead up to the proposed code changes in this PR. These commits are not part of the code changes. They are merely there for future reference and to provide a step-by-step guide on how to reproduce the issue as well as proof of the underlying cause before finally proposing a fix that provides a remedy to the risks listed in the ticket.

Therefore, it's important that the code changes are reviewed in the sequence the commit arrive before reviewing the proposed code changes described by this PR. In order:

1. [add test to reproduce defect](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/892/changes/f05094806535c623d11892d48e573a2a7f9a1097)

This test was written first to reproduce and demonstrate the issue as identified in the ticket. When the test is run in this commit, it indeed fails.

failure

2. [proof that serialisation fixes the defect as described in the test](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/892/changes/a102d8ba0faa5fca3d099e9b52f686b4bcdf04e1)

The changes in this commit are meant to demonstrate that the underlying cause is indeed fixed by serialising execution. When the test is run in this commit, it passes [^1].

serialised

[^1]: There is a caveat as described in the **Important Note** found in the commit message.

3. [throttle user session evaluation](https://github.com/govuk-one-login/mobile-ios-one-login-app/commit/ba6f04264c0aee778825de7cb6742845dbb17ae5)

This commit, in addition to serialising also throttles the user evaluation. Thus the test changes to now assert that in case multiple Task(s) are scheduled in quick succession, only 1 app session state will be reported.

The comments in each commit provide more context. Each commit can be independently checked out and verified.

# Testing

While the above test **demonstrates the issue** and asserts the proposed fix, it is not proof that the code has not regressed in any other way in relation to the user evaluation.

In order to test that, the following user scenario was exercised. 

> A user who is already authenticated (i.e. their user session is evaluated as `.saved`) is able to bring the app into the foreground over a period of time and see the home screen

That scenario must satisfy the following product requirements:
* A user will land on the home screen without encountering any errors
* The user session is evaluated as timely as possibly

While exercising the scenario, Instruments was used to monitor the creation, scheduling an execution of tasks while simultaneously a recording of the actual network calls made was captured to confirm the number of requests made as part of resuming a session (and by extension refreshing a token).

Below you can find a video that demonstrates the user behaviour, a timeline of the outbound network calls as well as a screenshot from Instruments of the Swift Tasks, both before and after the proposed code changes in this PR.

## Before

https://github.com/user-attachments/assets/bb8b1855-71ee-4c9b-8711-49361def1c55
Screenshot 2026-07-31 at 17 14 34

## After

https://github.com/user-attachments/assets/e72488c8-1686-49d1-bf56-874a989f555b
Screenshot 2026-07-31 at 17 14 44

# Where to focus the review

* Is the issue correctly identified in the ticket?
* Does the test correctly reproduce the issue? Is the test driving the code in an invalid way, makes the wrong assumptions or is misconstructed?
* Does the solution fix the issue without any side effects? Does the test coverage need to be expanded, how?

# Checklist

### DIFF
--- a/Sources/Qualification/State/AppQualifyingService.swift
+++ b/Sources/Qualification/State/AppQualifyingService.swift
@@ -7,7 +7,35 @@ import SecureStore
 
 @MainActor
 protocol QualifyingService: AnyObject {
+    /// Assign a delegate to receive state updates on:
+    /// * ``AppInformationState``
+    /// * ``AppSessionState``
+    /// * ``RemoteServiceState``
+    ///
+    /// Assigning a delegate guarantees that any future state updates will be delivered.
+    /// Past state updates, including the current state are NOT expected to delivered at the time a delegate is assigned.
+    ///
+    /// An implementation of ``QualifyingService`` chooses how and when state updates are delivered,
+    /// including whether they are delivered as they arrive or scheduled to be delivered (e.g. at the next opportunity)
+    ///
+    /// - SeeAlso: ``AppQualifyingService``
+    /// - Remark: Your implementation of ``QualifyingService`` must provide strong guarantees that state updates
+    /// are delivered in the "correct" sequence and represent valid transitions.
     var delegate: AppQualifyingServiceDelegate? { get set }
+    
+    /// Schedules an evaluation to both the ``AppInformationState`` and ``AppSessionState``.
+    /// Assign a ``delegate`` prior to the call to ``initiate()`` to receive state updates as they arrive.
+    ///
+    /// Once an evaluation is in flight, no further evaluations can be scheduled. In other words, multiple calls
+    /// to ``initiate()`` in quick succession are effectively no-op.
+    ///
+    /// This is by design as it's meant to avoid excessive, unnecessary evaluations that can effectively keep the system
+    /// busy and starve it off its ability to execute any other work.
+    ///
+    /// A state update is guaranteed to be delivered in the correct sequence and represent valid transitions
+    /// between states over time. While multiple calls to ``initiate()`` in quick succession do not result in
+    /// an equal number of evaluations, you are guaranteed to receive a state update in the correct sequence and represent valid transitions between states over time.
+    ///
     func initiate()
     func evaluateUserSession() async
 }
@@ -49,8 +77,8 @@ final class AppQualifyingService: QualifyingService {
             }
         }
     }
-
-    let serialTaskQueue: SerialTaskQueue = SerialTaskQueue()
+    
+    private var task: Task<Void, Never>?
 
     init(
         analyticsService: OneLoginAnalyticsService,
@@ -63,11 +91,16 @@ final class AppQualifyingService: QualifyingService {
     }
     
     public func initiate() {
-        Task {
-            try await serialTaskQueue.enqueue {
-                await self.qualifyAppVersion()
-                await self.evaluateUserSession()
+        guard task == nil else {
+            return
+        }
+
+        self.task = Task(name: #function) {
+            defer {
+                self.task = nil
             }
+            await qualifyAppVersion()
+            await evaluateUserSession()
         }
     }
     
--- a/Sources/Qualification/State/AppQualifyingService.swift
+++ b/Sources/Qualification/State/AppQualifyingService.swift
@@ -56,12 +56,10 @@ final class AppQualifyingService: QualifyingService {
     
     private var initiateTask: Task<Void, Never>?
     
-    /** used for testing only */
-    // swiftlint:disable identifier_name
+    /// used for testing only
     var _initiateTask: Task<Void, Never>? {
         initiateTask
     }
-    // swiftlint:enable identifier_name
 
     private var appInfoState: AppInformationState = .notChecked {
         didSet {
--- a/Sources/Qualification/State/AppQualifyingService.swift
+++ b/Sources/Qualification/State/AppQualifyingService.swift
@@ -54,6 +54,8 @@ final class AppQualifyingService: QualifyingService {
     private let sessionManager: SessionManager
     weak var delegate: AppQualifyingServiceDelegate?
     
+    private var initiateTask: Task<Void, Never>?
+
     private var appInfoState: AppInformationState = .notChecked {
         didSet {
             Task { [appInfoState = appInfoState] in
@@ -78,8 +80,6 @@ final class AppQualifyingService: QualifyingService {
         }
     }
     
-    private var task: Task<Void, Never>?
-
     init(
         analyticsService: OneLoginAnalyticsService,
         updateService: AppInformationProvider = AppInformationService(baseURL: AppEnvironment.appInfoURL),
@@ -91,13 +91,13 @@ final class AppQualifyingService: QualifyingService {
     }
     
     public func initiate() {
-        guard task == nil else {
+        guard initiateTask == nil else {
             return
         }
 
-        self.task = Task(name: #function) {
+        self.initiateTask = Task(name: #function) {
             defer {
-                self.task = nil
+                self.initiateTask = nil
             }
             await qualifyAppVersion()
             await evaluateUserSession()
--- a/Sources/Qualification/State/AppQualifyingService.swift
+++ b/Sources/Qualification/State/AppQualifyingService.swift
@@ -57,9 +57,11 @@ final class AppQualifyingService: QualifyingService {
     private var initiateTask: Task<Void, Never>?
     
     /** used for testing only */
+    // swiftlint:disable identifier_name
     var _initiateTask: Task<Void, Never>? {
         initiateTask
     }
+    // swiftlint:enable identifier_name
 
     private var appInfoState: AppInformationState = .notChecked {
         didSet {
--- a/Sources/Qualification/State/AppQualifyingService.swift
+++ b/Sources/Qualification/State/AppQualifyingService.swift
@@ -55,6 +55,11 @@ final class AppQualifyingService: QualifyingService {
     weak var delegate: AppQualifyingServiceDelegate?
     
     private var initiateTask: Task<Void, Never>?
+    
+    /** used for testing only */
+    var _initiateTask: Task<Void, Never>? {
+        initiateTask
+    }
 
     private var appInfoState: AppInformationState = .notChecked {
         didSet {
--- a/Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
+++ b/Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
@@ -563,27 +563,27 @@ extension AppQualifyingServiceTests {
         let appInformationStates: [AppInformationState] = [
             .qualified
         ]
-        let sessionStatesReceivedExpectation = expectation(description: "expected session states received")
-        sessionStatesReceivedExpectation.assertForOverFulfill = false
+        let appInformationStatesReceivedExpectation = expectation(description: "expected app information states received")
+        appInformationStatesReceivedExpectation.assertForOverFulfill = false
         let sut: AppQualifyingService = .make(appInformationProvider: MockAppInfoAppInformationProvider(appInfoStates: appInformationStates))
-        var receivedSessionStates = [AppInformationState]()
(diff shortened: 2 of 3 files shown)


Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -59,9 +61,14 @@ final class AppQualifyingService: QualifyingService {
...
     public func initiate() {
-        Task {
+        guard task == nil else {
+            return
+        }
+
+        self.task = Task(name: #function) {
             await qualifyAppVersion()
             await evaluateUserSession()
+            self.task = nil
```
chionia-gds: hypothetically, if a user gets stuck in something in qualifyAppVersion or evaluateUserSession and this line is not called, could this mean that they can be stuck and they would need to dismiss the app and restart?

zzcgumn: I think so.

qnoid: That's a good observation @chionia-gds. I was meant to make one more change then I forgot!

In short, no. I don't see how the user "gets stuck in something". Do you see such use case?

The change I wanted to make is to ensure the assigned task to nil is "always" called, by putting it in a defer block, in case **in the future** the code in `qualifyAppVersion` or `evaluateUserSession` changes to throw an error.

qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890/commits/7e3e67f1d4d75c9b4809d10d40b98bf6e73a07a0

Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -47,6 +47,8 @@ final class AppQualifyingService: QualifyingService {
             }
         }
     }
+    
+    private var task: Task<Void, Error>?
```
zzcgumn: Did you consider refactoring app qualifying service as an actor? I believe you would be able to replace the delegate callbacks with await on properties by changing from a final class to an actor.

As is, this is not `Sendable` and you would have to either mark the entire class as `@MainActor` or refactor it to be an actor to compile with strict concurrency enabled.

robinphillips: Btw, @qnoid - this is a great investigation, congratulations. 

Re your point @zzcgumn, looking at the One Login code, the `initiate()` function is called from a scene delegate method which is not `async` so somewhere has to encapsulate the async call chain and this seems like a reasonable place to me. 

One thing that might be helpful that we have done a few times for these kinds of functions that sit at boundaries of asynchronous code is to return the `Task` as an optional and add `@discardableResult` - this allows optionally awaiting the `value` such that tests can be more deterministic. Not a need, but sharing here in case it is helpful :)

qnoid: > return the Task as an optional and add @discardableResult - this allows optionally awaiting the value such that tests can be more deterministic. 

That is very useful @robinphillips. Thank you for bringing awareness to it.

qnoid: @zzcgumn to be honest I did not consider it! Generally speaking I prefer to keep changes to a minimum let alone refactor.

IMO the `AppQualifyingService`, `PersistentSessionManager` and their relationship could use some TLC. I will definitely consider the use of an actor in the future.

Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -59,7 +61,14 @@ final class AppQualifyingService: QualifyingService {
...
     
     public func initiate() {
-        Task {
+        guard task == nil else {
+            return
+        }
+
+        self.task = Task(name: #function) {
+            defer {
+                self.task = nil
```
robinphillips: Before the defer, is it also worth considering having a guard against the existing task being not nil? Eg:

```
guard task == nil else { return }
```

I think (and I could be wrong) that as things stand if `initiate()` is called multiple times in quick succession and the task hasn't completed yet (and therefore `defer` hasn't been called) that there's a chance of calling:

```
await qualifyAppVersion()
await evaluateUserSession()
```
again in a separate task. Maybe another option would be to handle task cancellation such that before assigning the new task, you cancel the old one but that might be undesirable.

qnoid: The guard does exist as a precondition to the initiate.

```
        guard task == nil else {
            return
        }
```

Effectively, if and only if the Task is nil the code will create a new one. Since the Task sets the reference to nil once it has completed, it's not possible to call `initiate()` in quick succession which will result in multiple Tasks. I was under the impression that the `test_multiple_foreground_initiate_calls_do_not_lead_to_invalid_session_state_transition` test proves that but maybe not? 🤔

> Maybe another option would be to handle task cancellation such that before assigning the new task, you cancel the old one but that might be undesirable.

This is not desirable. Replacing and cancelling the existing task is a different policy. That would effectively mean every time the app comes into the foreground, the previous evaluation would be dismissed and start anew. The old task might still be in the middle of mutating state since task cancellation is cooperative. 

The proposed policy ignores duplicate requests until the previous one has finished and  ensures that execution is serialised so only one task ever executes before another one operates on shared data.

robinphillips: Sorry, must have missed the guard that's there - think its fine :)

Sources/Qualification/State/AppQualifyingService.swift
qnoid: @zzcgumn to be honest, I would propose the `AppQualifyingService` is annotated as MainActor rather than make it a separate actor. I expect that would also fix the [known issue](https://govukverify.atlassian.net/browse/DCMAW-21050) for `appInfoState`, `sessionState`, `serviceState` by removing the Task/await.

qnoid:
On the back of @zzcgumn and @chionia-gds feedback, I reviewed the changes and I can see a race condition between the main thread (i.e. MainActor from `SceneDelegate`) calling `initiate()` and the background thread where the Task runs.

This is due to the fact that neither the `AppQualifyingService` or the `initiate()` function are annotated with an actor (e.g. MainActor). This ambiguity leaves room for:
* Potentially the call `initiate()` to run on any actor (i.e. thread)
* The Task in `initiate()` to run its closure on any actor (i.e. thread)

Effectively, you can have a race between:
* the "initiate" thread setting the value of `self.task` to `Task` 
* the `Task` thread setting the value of `self.task` to `nil`

If the `Task` thread wins the race, then
* `self.task = nil`

followed by 

* `self.task = Task`

Any followup call to `initiate` will evaluate `guard task == nil` to false, preventing any future user session evaluations.

The solution to this is to annotate `initiate` with `@MainActor`. In this case, the [Task inherits MainActor isolation](https://github.com/swiftlang/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#_inheritactorcontext) which effectively means that the operation is guaranteed to run after the assignment of task, `self.task = Task`, once the MainActor has been freed to execute the operation.

With that in mind and the [known defect](https://govukverify.atlassian.net/browse/DCMAW-21050), an [all encompassing solution would be to annotate AppQualifyingService with MainActor](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890#discussion_r3695448926).

qnoid:
There is now a [separate PR](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/892) that changes the `AppQualifyingService` to be annotated with `@MainActor`. Once merged, I will rebase this PR which will no longer be affected by the race condition.

qnoid:
On a separate note, I would like to see if we can expand the test coverage of this PR to ensure that:
1. multiple calls on initiate in quick succession only result in 1 evaluation.
2. a second call once the last evaluation is completed, will result in a new evaluation.

For the first test, the only way to write a test for the use case to assert that initiate() only ever results in 1 callback to the delegate when multiple calls are made, is to get a reference to the Task and await its value before asserting on the received application states.

However, given how the Task must be encapsulated and not allowed to escape, one solution I see is to **add another function that returns the Task and is only available to unit tests**. Since there is no access modifier to only make a function available to testing, how do we feel about adding a conditional compilation flag (e.g. TESTING) that is only defined in the test target (similar but not the same as to how we define the IS_RUNNING_TESTS environment variable)?

With a conditional compilation flag, we can wrap the "test function" in that TESTING , similar to DEBUG (e.g. `#if TESTING`) so that it is only compiled in the source when tests are running, thus making it visible to unit tests but not part of the public or module interface.

Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -47,6 +47,8 @@ final class AppQualifyingService: QualifyingService {
             }
         }
     }
+    
+    private var task: Task<Void, Never>?
```
shukriahmed-dd: just a small nit: instead of calling this `task` how about `initiateTask` or `initiationTask` ? so its more readable. Also i wonder should this be defined on line 26 instead of down here?

shukriahmed-dd (approved):
looks good ! thank you for all the notes / ticket details & order of commits! it was easy to understand the problem and the fix :)

qnoid:
> looks good ! thank you for all the notes / ticket details & order of commits! it was easy to understand the problem and the fix :)

I am happy to see it does not go unnoticed, is helpful and appreciated! Thank you!

Sources/Qualification/State/AppQualifyingService.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890/changes/8216ab5f95b09c943ec54dff3c22af883430a77c

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
```diff
@@ -460,31 +550,125 @@ extension AppQualifyingServiceTests {
...
-    /// Since the ``AppQualifyingService`` **schedules a Task** to perform a callback to the delegate, it's possible that
-    /// by the time the task runs and reads the value, the value has since change from another call to ``qualifyAppVersion``
+    /// - SeeAlso: ``SceneDelegate/sceneWillEnterForeground(_:)``
+    /// - SeeAlso: ``AppQualifyingService/appInfoState``
+    /// - SeeAlso: ``MockAppInfoAppInformationProvider``
+    func test_multiple_foreground_initiate_calls_only_evaluate_information_state_once() async throws {
+        let appInformationStates: [AppInformationState] = [
+            .qualified
+        ]
+        let sessionStatesReceivedExpectation = expectation(description: "expected session states received")
```
jamielutton-gds: This test refers to "session states" in a few different places, like this variable `sessionStatesReceivedExpectation` and its description (`expected session states received`) or the variable a few lines down `receivedSessionStates`, but it looks like these should instead be referring to "app information states" based on the setup [here](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890/changes#diff-5b661bc566971a206f5cffa7a95b44fbd199b4f451065b443996173b3504b37bR563) which initialises `appInformationStates` and the assertion [here](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890/changes#diff-5b661bc566971a206f5cffa7a95b44fbd199b4f451065b443996173b3504b37bR600) which validates `receivedSessionStates` and `appInformationStates` are equal (which doesn't feel quite right, comparing "session states" to "app information states"). Can you check that all of the variables in this test and the one below (`test_foreground_initiate_in_sequence_evaluate_every_information_state`) are appropriately named for the case where app information states are evaluated?

jamielutton-gds (approved):

PR #892, opened 2026-08-03
feat: DCMAW-21050 ensure delegate callback does not misrepresent state

qnoid:
# [iOS | Tech Debt | A delegate callback on the sessionState, appInfoState and (possibly) the serviceState is sensitive to timings and may report an unexpected/wrong state.](https://govukverify.atlassian.net/browse/DCMAW-21050)

The proposed code changes ensure `AppQualifyingService` reliably reports on `AppInformationState`, `AppSessionState` and `RemoteServiceState` state and transitions.

# Changes
The fix requires 2 changes:

1. Calls on `qualifyAppVersion` and `evaluateUserSession` across different scheduled Task(s) due to multiple calls to `initiate()` over time is serialised

2. The callback on to report on each state `AppInformationState`, `AppSessionState`, `RemoteServiceState` is done with the value that has just been written at the time the Task is created instead of the value it has when its executed.

# Testing
[One commit](https://github.com/govuk-one-login/mobile-ios-one-login-app/commit/f05094806535c623d11892d48e573a2a7f9a1097), adds a test to reproduce defect:

Screenshot 2026-08-03 at 14 27 11
Screenshot 2026-08-03 at 14 27 19

[Second commit](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/892/changes/a102d8ba0faa5fca3d099e9b52f686b4bcdf04e1) with the code changes that fixes the issue. Each test run for 100 iterations to have high confidence there is no timing at play:

Screenshot 2026-08-03 at 14 54 55
Screenshot 2026-08-03 at 14 55 17

# Checklist

### DIFF
--- a/Sources/Qualification/State/AppQualifyingService.swift
+++ b/Sources/Qualification/State/AppQualifyingService.swift
@@ -5,6 +5,7 @@ import MobilePlatformServices
 import Networking
 import SecureStore
 
+@MainActor
 protocol QualifyingService: AnyObject {
     var delegate: AppQualifyingServiceDelegate? { get set }
     func initiate()
@@ -18,6 +19,7 @@ protocol AppQualifyingServiceDelegate: AnyObject {
     func didChangeServiceState(state: RemoteServiceState)
 }
 
+@MainActor
 final class AppQualifyingService: QualifyingService {
     private let analyticsService: OneLoginAnalyticsService
     private let updateService: AppInformationProvider
@@ -26,27 +28,29 @@ final class AppQualifyingService: QualifyingService {
     
     private var appInfoState: AppInformationState = .notChecked {
         didSet {
-            Task {
-                await delegate?.didChangeAppInfoState(state: appInfoState)
+            Task { [appInfoState = appInfoState] in
+                delegate?.didChangeAppInfoState(state: appInfoState)
             }
         }
     }
     
     private var sessionState: AppSessionState = .notLoggedIn {
         didSet {
-            Task {
-                await delegate?.didChangeSessionState(state: sessionState)
+            Task { [sessionState = sessionState] in
+                delegate?.didChangeSessionState(state: sessionState)
             }
         }
     }
     
     private var serviceState: RemoteServiceState = .activeService {
         didSet {
-            Task {
-                await delegate?.didChangeServiceState(state: serviceState)
+            Task { [serviceState = serviceState] in
+                delegate?.didChangeServiceState(state: serviceState)
             }
         }
     }
+
+    let serialTaskQueue: SerialTaskQueue = SerialTaskQueue()
 
     init(
         analyticsService: OneLoginAnalyticsService,
@@ -60,8 +64,10 @@ final class AppQualifyingService: QualifyingService {
     
     public func initiate() {
         Task {
-            await qualifyAppVersion()
-            await evaluateUserSession()
+            try await serialTaskQueue.enqueue {
+                await self.qualifyAppVersion()
+                await self.evaluateUserSession()
+            }
         }
     }
     
@@ -95,7 +101,6 @@ final class AppQualifyingService: QualifyingService {
         }
     }
     
-    @MainActor
     func evaluateUserSession() async {
         guard appInfoState == .qualified else {
             // Do not continue with local auth unless app info qualifies
--- a/Tests/UnitTests/Mocks/Sessions+Services/Storage/MockSessionManager.swift
+++ b/Tests/UnitTests/Mocks/Sessions+Services/Storage/MockSessionManager.swift
@@ -237,3 +237,76 @@ class MockSessionManagerExpectation: SessionManager {
         try await sessionManager.clearAppForLogin()
     }
 }
+
+/// A mock that emulates the fact that ``sessionState`` can change over time similar to the implementation of
+/// ``PersistentSessionManager``, including while waiting for``resumeSession`` to return.
+///
+/// Specifically, this mock:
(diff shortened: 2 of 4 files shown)


Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -60,8 +64,10 @@ final class AppQualifyingService: QualifyingService {
     
     public func initiate() {
         Task {
-            await qualifyAppVersion()
-            await evaluateUserSession()
+            try await serialTaskQueue.enqueue {
```
zzcgumn, line 67: I think we should document that calling the `initiate` method twice fires two tasks executed serially. Perhaps also what happens with the second task, memory serves me correctly this will be different from the first.

zzcgumn (approved):
This is perfectly fine as long as we are happy with the design decision to use fire and forget delegate calls.

The alternative, as I see it, would be to publish property updates through an AsyncSequence. This can make it easier to verify the code at the subscribing side, but does not reduce complexity compared to the delegate pattern.

Sources/Qualification/State/AppQualifyingService.swift
```diff
@@ -5,6 +5,7 @@ import MobilePlatformServices
 import Networking
 import SecureStore
 
+@MainActor
 protocol QualifyingService: AnyObject {
     var delegate: AppQualifyingServiceDelegate? { get set }
```
zzcgumn, line 10: Do we send the current state when a delegate is assigned?

Sources/Qualification/State/AppQualifyingService.swift
qnoid: I see this [implementation as a stopgap](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890/changes/e9d04a8c8c24cffd68f87b9849fef120303b5bdd).
It will change once again with https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890, which will drop the need to use the `SerialTaskQueue`. Is it ok with you to document the `initiate()` in that PR?

qnoid:
> This is perfectly fine as long as we are happy with the design decision to use fire and forget delegate calls.
> 
> The alternative, as I see it, would be to publish property updates through an AsyncSequence. This can make it easier to verify the code at the subscribing side, but does not reduce complexity compared to the delegate pattern.

That's a great shout. I can see how `AsyncSequence` is better suited to communicate a "stream of state updates" which ensures serialised delivery. I will keep in in mind for the future. Thank you.

Sources/Qualification/State/AppQualifyingService.swift
qnoid: It does not look like the code was doing this prior and this PR does not change that. Are you pointing out to the fact that it should?

Tests/UnitTests/Mocks/Sessions+Services/Storage/MockSessionManager.swift
```diff
@@ -237,3 +237,76 @@ class MockSessionManagerExpectation: SessionManager {
...
+/// Specifically, this mock:
+/// * Dispenses a new ``sessionState`` each time one is requested to emulate state changes over time
+/// * Adds a **random delay** on a call to ``resumeSession`` to emulate the variability and delay of a network response
+
+/// Use this mock to write a test that will fail **should the code under test detects an unexpected state transition that can arise
+/// when waiting for ``resumeSession``**, in adition to the test assertions.
+///
+/// You create an instance of this mock using ``init(sessionStates:)`` by passing a list of `sessionStates` that will
+/// be dispensed in order first in; first out.
+final class MockResumeSessionSessionManager: SessionManager {
```
chionia-gds, line 253: is it necessary that we make a new mock? I feel like it's really hard to keep up with all the mocks for the same service, we already have MockSessionManager and MockSessionManagerExpectation. Can either of them be utilised or updated for this purpose?

qnoid: I believe the `MockSessionManager` and `MockSessionManagerExpectation` could be collapsed into one.

This one on the other hand has a very specific purpose.

chionia-gds (approved):

Sources/Qualification/State/AppQualifyingService.swift
zzcgumn: I don't think there is a need to change this if this is how it worked. What I think we should do is document the behaviour.

Sources/Qualification/State/AppQualifyingService.swift
zzcgumn: Yes, perfectly fine.

shukriahmed-dd (approved):

Sources/Qualification/State/AppQualifyingService.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/6a0d4106e2d1b5ea6ab35e2f3c532a811e19a936/Sources/Qualification/State/AppQualifyingService.swift#L26-L39 in https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890

Sources/Qualification/State/AppQualifyingService.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/6a0d4106e2d1b5ea6ab35e2f3c532a811e19a936/Sources/Qualification/State/AppQualifyingService.swift#L10-L24 in https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/890

PR #912, opened 2026-09-01
feat: DCMAW-20314 adopt WalletSDK/delete

qnoid:
# [DCMAW-20314 adopt WalletSDK/delete](https://govukverify.atlassian.net/browse/DCMAW-20314)

This code change migrates away from the deprecated [WalletSDK/deleteData()](https://github.com/govuk-one-login/mobile-wallet-ios/blob/18.72.1/Wallet/Sources/Wallet/WalletSDK.swift#L82) to the suggested [WalletSDK/delete](https://github.com/govuk-one-login/mobile-wallet-ios/blob/18.72.1/Wallet/Sources/Wallet/WalletSDK.swift#L93).

It also changes the `WalletSessionData` type so that it's possible to stream warnings emitted by the `WalletSDK/delete` and are intended to be "informational" only (e.g. logged for observability)

Finally, the warnings are logged in the analytics service.

# Testing

https://github.com/user-attachments/assets/fa9fb62e-8a2c-4f46-9429-6452a3568809

# Noteworthy
The `WalletSDKWrapper` now defines an `instance` property. This is meant to indicate that this is meant to be treated as a  singleton that represents the `WalletSDK` and the functionality is scoped only to its static functions.

# Scout rule
* Made `LocalAuthManaging` injectable when making a `PersistentSessionManager`
* Updated the `PersistentSessionManager` instance under test to use the `.make()` function. This ensures that the instance of the `PersistentSessionManager` under test is indeed the one used by the One Login codebase.
* * That change was made in response to discovering that the `test_saveSession_doesNotRefreshSecureStoreManager` test was a false positive since it didn't test the `PersistentSessionManager` as returned by `make()`.

# Checklist

### DIFF
--- a/Sources/Service/WalletSessionData+OneLogin.swift
+++ b/Sources/Service/WalletSessionData+OneLogin.swift
@@ -1,16 +1,152 @@
 import Wallet
+import WalletStore
 
-struct WalletSessionData: SessionBoundData {
+/// Use to clear session data related to Wallet.
+///
+/// - SeeAlso: ``clearSessionData()`` for how to clear the session data.
+/// - SeeAlso: ``streamClearSessionDataWarnings`` on how to best consume a stream of warnings.
+actor WalletSessionData: SessionBoundData {
+    
+    /// Use to stream warnings emitted by each call to ``clearSessionData()``.
+    ///
+    /// To receive the warnings emitted by a call to ``clearSessionData()``, obtain the warning stream first
+    /// followed by a call to ``clearSessionData()``.
+    ///
+    /// There are two recommended ways to consume the stream, in order of preference:
+    /// 1. As a structured task, using `async let`.
+    /// 2. As a `Task` scoped to observing the warnings.
+    ///
+    /// For example, using a structured task.
+    ///
+    /// ```
+    /// func clearSessionData(
+    ///     walletSessionData: WalletSessionData
+    /// ) async throws {
+    ///     let warnings = await walletSessionData.streamClearSessionDataWarnings
+    ///
+    ///     async let logWarnings: Void = {
+    ///         for await warning in warnings {
+    ///             analyticsService.logCrash(warning)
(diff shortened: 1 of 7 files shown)


Sources/Service/WalletSessionData+OneLogin.swift
```diff
@@ -1,16 +1,152 @@
...
+    /// ```
+    ///
+    /// You **must** wait for ``clearSessionData()`` to complete before obtaining another stream;
+    /// overlapping calls are not supported and may result in warnings being lost.
+    ///
+    /// - SeeAlso: ``clearSessionData()`` for how warnings are produced.
+    ///
+    /// - Note: If a ``WalletSessionData`` instance is unexpectedly retained, a consumer waiting
+    /// on a stream will remain suspended in the case of either omitting a call to ``clearSessionData()``
+    /// or the observation task is not cancelled.
```
shukriahmed-dd, line 70: helpful comments, I wonder if its better to have these technical notes in confluence instead of in the source code, maybe a folder can be created here https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3795320911/Mobile+Tech+Content?atlOrigin=eyJpIjoiNTJlMGY3ZDlmMzgwNDMyZWI2NjBkMjU0ODkwMDhiYTEiLCJwIjoiYyJ9

shukriahmed-dd (approved):

chionia-gds (approved):

Sources/Service/WalletSessionData+OneLogin.swift
qnoid: In general I agree. Some technical notes could go either into Confluence or in a README or in a Tech Design document within the repo. Say a fundamental choice or behaviour that one should know going into the codebase.  I don't think this is the case here? This is how this code works and as a developer how to use it.

PR #914, opened 2026-09-03
feat: DCMAW-22079 persistent session id is missing for a returning user

qnoid:
# [iOS | Tech Debt | SecureStoreError(.cantDecryptData) | NSOStatusErrorDomain (-50) | PersistentSessionID is missing for a returning user](https://govukverify.atlassian.net/browse/DCMAW-22079)

  This is case where attempting to start an auth session, throws a `PersistentSessionError(.sessionMismatch)` error.
  In this case, the user has restored from a backup and the `persistentID` is not accessible (e.g. cannot be decrypted)
  due to the fact that the private key backed by the SE is not exportable.

  This code change adds an assertion to verify that a returning user can indeed login,
  without encountering a `PersistentSessionError(.sessionMismatch)` **in case they have restored from a backup**.
  The `PersistentSessionError(.sessionMismatch)` may still be thrown in other cases[1].

  In case the assertion fails, due to a `SecureStoreError(.cantDecryptData)` error,
  all of user data is deleted and the user is presented with the login screen as normal, which is expected to succeed.

  In case the assertion fails due to any other error, the session is considered to have failed.

  The assertion happens while the loading screen is shown, before attempting to evaluate the
  user session so that the user is either shown the login or an unrecoverable screen in case an error is thrown.

  [1]: At this moment, there are no other known cases. The `PersistentSessionError(.sessionMismatch)`
  aka `PersistentSessionErrorKind (1003)` has been [seen in our logs in the thousands](https://console.firebase.google.com/project/mobile-one-login-prod/crashlytics/app/ios:uk.gov.onelogin/issues/338a4fb1d5b8d1951e3cb2abf61e2bbd?time=90d&sessionEventKey=d2f8c23e103945fda177a843791c802d_2259759493317190388).
  This code change aims to draw a distinction between a “legitimate” `PersistentSessionError(.sessionMismatch)`
  which may indicate a programmatic error and a `PersistentSessionError(.sessionMismatch)` due to a `SecureStoreError(.cantDecryptData)` error (e.g. [as seen on staging](https://console.firebase.google.com/project/mobile-one-login-staging/crashlytics/app/ios:uk.gov.onelogin.staging/issues/5843b557308416b1e186ba18b1556b59?time=7d&sessionEventKey=9ceb4799bb4f45799e65e1d46ea0f591_2259808495129907350) while testing the change in this PR).

# Changes

## Product

**In case of a restore**
Previously, once a returning user has restored from a backup, two scenarios were possible:

## Before

before

1. The access token has expired. In which case the user would be asked to sign in again. This attempts to [clear all user data](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.29.0/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L218) and [it throws a](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.29.0/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L223) `PersistentSessionError(.sessionMismatch)`. The `PersistentSessionError` [is handled](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Login/LoginCoordinator.swift#L99), by showing a [data delete warning screen](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Login/LoginCoordinator.swift#L231), allowing the user to login again.

login before

https://github.com/user-attachments/assets/24eb5bd9-b2db-4b07-ba18-598a83b66362

2. The access token is valid. In which case the user would be asked to resume their session. Since `persistentId` is inaccessible, a `PersistentSessionError(.noSessionExists)` [is thrown](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.29.0/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L298). That attempts to [clear all the user session data](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Qualification/State/AppQualifyingService.swift#L189). A `.systemLogOut` notification is posted, which [publishes an update to the sessionState](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Qualification/State/AppQualifyingService.swift#L239), [which displays the login coordinator](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Qualification/QualifyingCoordinator.swift#L101) which [already exists](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Qualification/QualifyingCoordinator.swift#L134) hence it displays its "root", without the data deleted warning screen, which is the "You need to sign in again" screen.

resume before

https://github.com/user-attachments/assets/91e3a69e-67c5-4a53-af8e-4b8df619194d

With these changes, before attempting to evaluate the session, [a call is made to assertReturningUserCanLogin](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/changes#diff-2b8361fc4b906f97e8fc9aed97a2b0f429b06eccfaf9ea47a6c33fbb7adb0a91R149). [In case the data cannot be decrypted](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/changes#diff-d9e901b931fb93a0c3733e4284970bf1a470ddf02ad8e7985adf02e7cbd03299R270), the user session data is deleted and the `SecureStoreError(.cantDecryptData)` error [is thrown](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/changes#diff-d9e901b931fb93a0c3733e4284970bf1a470ddf02ad8e7985adf02e7cbd03299R273). A `.systemLogOut` notification is posted, which [publishes an update to the sessionState](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/741c1b95a29466a60ec6843cede67c0907de85eb/Sources/Qualification/State/AppQualifyingService.swift#L251), [which launches the login coordinator](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/741c1b95a29466a60ec6843cede67c0907de85eb/Sources/Qualification/QualifyingCoordinator.swift#L101) and [presents the data deleted warning screen.](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/741c1b95a29466a60ec6843cede67c0907de85eb/Sources/Login/LoginCoordinator.swift#L158).

## After
after

https://github.com/user-attachments/assets/1361c6f8-15b6-4a9a-88c3-3b8ecaad1b76

# Do not merge

This PR depends on:
* https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/912

# Testing

https://github.com/user-attachments/assets/c47a1e6d-32a4-4307-b329-3452e7bec542

# Discussion

1. The access token has expired. In which case the user would be asked to sign in again.

Previously, [an attempt to login](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.29.0/Sources/Login/LoginCoordinator.swift#L97) that throws a `PersistentSessionError(.cannotDeleteData)` (i.e. [in case clear all session data failed](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/develop/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L221)), would show [the recoverable screen](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/1.29.0/Sources/Login/LoginCoordinator.swift#L233). This was [a case of user stuck in a login loop](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/833). In this case, this was a permanent state where the user had to delete the app to recover.

This behaviour is preserved. See https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/changes/398682a2d1c9d2b42c4c30c088e069e96f918f5f for details.

~With this code change, that will be treated as an unrecoverable error. AFAIK, out of all the session data, calling `clearSessionData` only the `WalletSessionData` [can throw](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/feat/DCMAW-22079-persistentSessionID-is-missing-for-a-returning-user/Sources/Service/WalletSessionData%2BOneLogin.swift#L117). In case it does, [the error is considered "critical"](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/feat/DCMAW-22079-persistentSessionID-is-missing-for-a-returning-user/Sources/Service/WalletSessionData%2BOneLogin.swift#L111). Is it safe to assume that the `PersistentSessionError(.cannotDeleteData)` is permanent/unrecoverable? If not, maintaining the existing behaviour of showing the recoverable screen, will put the user in a loop. On a **future note**, should there be a screen that lets the user know the only way to recover is to delete the app? Something to discuss with the wider team. 🤔~

before

# Related

* [Slack Thread](https://gds.slack.com/archives/C06UQBUJGP8/p1783681987773749?thread_ts=1783523501.298529&cid=C06UQBUJGP8)
* `PersistentSessionError(.noSessionExists)` (aka PersistentSessionErrorKind (1001) since the error code was standardised) [on Firebase, last 90 days.](https://console.firebase.google.com/project/mobile-one-login-prod/crashlytics/app/ios:uk.gov.onelogin/issues/5c252923926783abbb3598f7c33e57fc?time=90d&types=error&customKeys=kind%E2%87%92noSessionExists&sessionEventKey=a104c13135964dc891f8389fb0118dab_2260082873340820357)
* _Any_ PersistentSessionError(.noSessionExists), before the error was was standardised on [Firebase, last 90 days.](https://console.firebase.google.com/project/mobile-one-login-prod/crashlytics/app/ios:uk.gov.onelogin/issues?time=90d&state=all&tag=all&sort=eventCount&issuesQuery=key:kind%20value:OneLogin.PersistentSessionErrorKind.noSessionExists&types=error)

# Checklist

### DIFF
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -202,7 +202,7 @@ final class PersistentSessionManager: SessionManager {
     }
     
     /// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
-    private func _clearAllSessionData(presentSystemLogOut: Bool = true) async throws {
+    private func clearAllSessionData() async throws {
         // I am a returning user
         // but cannot reauthenticate because I don't have a persistent session ID
         //
@@ -212,14 +212,14 @@ final class PersistentSessionManager: SessionManager {
                 analyticsService.logCrash(PersistentSessionError(.sessionMismatch,
                                                                  reason: "secure wallet data deleted"))
             }
-            try await clearAllSessionData(presentSystemLogOut: presentSystemLogOut)
+            try await clearAllSessionData(presentSystemLogOut: true)
         } catch {
             throw PersistentSessionError(.cannotDeleteData, originalError: error)
         }
     }
     
     /// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
-    private func _clearAppForLogin() async throws {
+    private func prepareAppForLogin() async throws {
         // I am a first time user
         // I don't have a persistent session ID
         //
@@ -246,10 +246,10 @@ final class PersistentSessionManager: SessionManager {
     private func assertSession() async throws {
         if persistentID == nil {
             if isReturningUser {
-                try await _clearAllSessionData()
+                try await clearAllSessionData()
                 throw PersistentSessionError(.sessionMismatch)
             } else {
-                try await _clearAppForLogin()
+                try await prepareAppForLogin()
             }
         }
     }
@@ -269,7 +269,7 @@ final class PersistentSessionManager: SessionManager {
             
             if case .failure(let error as SecureStoreError) = persistentID, error.kind == .cantDecryptData,
                let originalError = error.originalError as? NSError, originalError.code == errSecParam {
-                try await self._clearAllSessionData()
+                try await self.clearAllSessionData()
                 throw error
             }
             
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -201,43 +201,87 @@ final class PersistentSessionManager: SessionManager {
         (try? localAuthentication.canUseAnyLocalAuth) ?? false && isReturningUser
     }
     
+    /// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
+    private func _clearAllSessionData(presentSystemLogOut: Bool = true) async throws {
+        // I am a returning user
+        // but cannot reauthenticate because I don't have a persistent session ID
+        //
+        // I need to delete my session & Wallet data before I can login
+        do {
+            if await !walletSDK.isEmpty() {
+                analyticsService.logCrash(PersistentSessionError(.sessionMismatch,
+                                                                 reason: "secure wallet data deleted"))
+            }
+            try await clearAllSessionData(presentSystemLogOut: presentSystemLogOut)
+        } catch {
+            throw PersistentSessionError(.cannotDeleteData, originalError: error)
+        }
+    }
+    
+    /// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
+    private func _clearAppForLogin() async throws {
+        // I am a first time user
+        // I don't have a persistent session ID
+        //
+        // I need to delete my session (but not analytics permissions) & Wallet data before I can login
+        do {
+            if await !walletSDK.isEmpty() {
+                analyticsService.logCrash(PersistentSessionError(.noSessionExists,
+                                                                 reason: "secure wallet data deleted"))
+            }
+            try await clearAppForLogin()
+        } catch {
+            throw PersistentSessionError(.cannotDeleteData, originalError: error)
+        }
+    }
+    
+    /// Asserts that a ``persistentID`` is present and accessible for either a returning or a first time user.
+    ///
+    /// - throws: ``PersistentSessionError(.sessionMismatch)`` in case this is a returning user without a persistent session ID.
+    ///     All session data will be cleared and a ``.systemLogUserOut`` notification is posted.
+    /// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
+    /// - Note: in case of a first time user, their "preferences" are not cleared.
+    /// - SeeAlso: ``clearAllSessionData(presentSystemLogOut:)`` for a returning user
+    /// - SeeAlso: ``clearAppForLogin()`` for a first time user
+    private func assertSession() async throws {
+        if persistentID == nil {
+            if isReturningUser {
+                try await _clearAllSessionData()
+                throw PersistentSessionError(.sessionMismatch)
+            } else {
+                try await _clearAppForLogin()
+            }
+        }
+    }
+    
+    private var assertReturningUserCanLoginEvaluated = false
+    
+    public func assertReturningUserCanLogin() async throws {
+        try await self.serialTaskQueue.enqueue {
+            guard self.isReturningUser, !self.assertReturningUserCanLoginEvaluated else {
+                return
+            }
+            
+            let persistentID = Result {
+                try self.encryptedStore
+                    .readItem(itemName: OLString.persistentSessionID)
+            }
+            
+            if case .failure(let error as SecureStoreError) = persistentID, error.kind == .cantDecryptData,
+               let originalError = error.originalError as? NSError, originalError.code == errSecParam {
+                try await self._clearAllSessionData()
+                throw error
+            }
+            
+            self.assertReturningUserCanLoginEvaluated = true
+        }
+    }
+    
     func startAuthSession(
         _ session: any LoginSession,
         using configuration: @Sendable (String?) async throws -> LoginSessionConfiguration
     ) async throws {
-        if persistentID == nil {
-            if isReturningUser {
-                // I am a returning user
-                // but cannot reauthenticate because I don't have a persistent session ID
-                //
-                // I need to delete my session & Wallet data before I can login
-                do {
-                    if await !walletSDK.isEmpty() {
-                        analyticsService.logCrash(PersistentSessionError(.sessionMismatch,
-                                                                         reason: "secure wallet data deleted"))
-                    }
-                    try await clearAllSessionData(presentSystemLogOut: true)
-                } catch {
-                    throw PersistentSessionError(.cannotDeleteData, originalError: error)
-                }
-                
-                throw PersistentSessionError(.sessionMismatch)
-            } else {
-                // I am a first time user
-                // I don't have a persistent session ID
-                //
-                // I need to delete my session (but not analytics permissions) & Wallet data before I can login
-                do {
-                    if await !walletSDK.isEmpty() {
-                        analyticsService.logCrash(PersistentSessionError(.noSessionExists,
-                                                                         reason: "secure wallet data deleted"))
-                    }
-                    try await clearAppForLogin()
-                } catch {
-                    throw PersistentSessionError(.cannotDeleteData, originalError: error)
-                }
-            }
-        }
+        try await assertSession()
         
         let response = try await session.performLoginFlow(
             configuration: configuration(persistentID)
--- a/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
+++ b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift
@@ -7,6 +7,7 @@ import GDSUtilities
 import LocalAuthenticationWrapper
 import Logging
 import SecureStore
+import WalletStore
 
 // swiftlint:disable:next type_body_length
 final class PersistentSessionManager: SessionManager {
@@ -486,6 +487,13 @@ public enum PersistentSessionErrorKind: Int, GDSErrorKind {
 }
 
 public typealias PersistentSessionError = OneLoginGDSError<PersistentSessionErrorKind>
+
+extension PersistentSessionError {
+    /// Returns true in case the underlying error is ``WalletStoreError(.walletUnsafeState)``.
+    var isWalletUnsafeState: Bool {
+        (originalError as? WalletStoreError)?.kind == .walletUnsafeState
+    }
+}
 
 protocol SessionBoundData {
     func clearSessionData() async throws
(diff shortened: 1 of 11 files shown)


chionia-gds (approved):

shukriahmed-dd (approved):

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
```diff
@@ -130,9 +132,236 @@ final class AppQualifyingServiceTests: XCTestCase {
...
+
+        #expect(sessionManager.didAssertReturningUserCanLogin)
+        #expect(expectedAppSessionState == nil)
+        #expect(sessionManager.sessionState == .expired)
+    }
+
+    @Test(
+        """
+        ON THE CONDITION a call to SessionManager/assertReturningUserCanLogin
+            throws a SecureStoreError(.cantDecryptData)
```
jamielutton-gds: Should this be "`throws a PersistentSessionError(.cannotDeleteData)`"?

That's what the test looks to set up [here](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/changes#diff-5b661bc566971a206f5cffa7a95b44fbd199b4f451065b443996173b3504b37bR191)

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
```diff
@@ -130,9 +132,236 @@ final class AppQualifyingServiceTests: XCTestCase {
...
+        GIVEN an AppQualifyingService with an `.expired` session state
+        WHEN the call to `AppQualifyingService/initiate` is finished
+        THEN SessionManager/assertCantDecryptData is called
```
jamielutton-gds: Should this be:
```suggestion
        THEN SessionManager/assertReturningUserCanLogin is called
```

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
```diff
@@ -130,9 +132,236 @@ final class AppQualifyingServiceTests: XCTestCase {
...
+    }
+    
+    @Test(
+        """
+        ON THE CONDITION a call to SessionManager/assertReturningUserCanLogin
+            throws a SecureStoreError(.cantDecryptData)
+        AND `originalError` is WalletStoreError(.walletUnsafeState)
+        GIVEN an AppQualifyingService with an `.expired` session state
+        WHEN the call to `AppQualifyingService/initiate` is finished
+        THEN SessionManager/assertCantDecryptData is called
```
jamielutton-gds: Should this be:
```suggestion
        THEN SessionManager/assertReturningUserCanLogin is called
```

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -201,43 +202,87 @@ final class PersistentSessionManager: SessionManager {
         (try? localAuthentication.canUseAnyLocalAuth) ?? false && isReturningUser
     }
     
-    func startAuthSession(
-        _ session: any LoginSession,
-        using configuration: @Sendable (String?) async throws -> LoginSessionConfiguration
-    ) async throws {
+    /// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
+    private func clearAllSessionData() async throws {
```
jamielutton-gds, line 206: Nitpick: Just thinking about logical groupings of similar functions - is there any benefit to defining this function next to the existing [clearAllSessionData function](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/07db32148b29fd68829f9185a06bc915fd203263/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L388), perhaps for ease comparison of their differences?

jamielutton-gds, line 206: Nitpick: In addition to this there is another [clearAllSessionData function](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/07db32148b29fd68829f9185a06bc915fd203263/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L388), with a different signature. Is it clear in what scenarios we'd use one over the other in future? 

This new function you've added does include comments about under what circumstances the function should be called, but it doesn't seem clear how you would know these things are true before calling the function, i.e. the comments say "I am a returning user but cannot reauthenticate because I don't have a persistent session ID" - that sounds like something that the function has determined, but it has in fact been determined outside of the function and is actually telling you about the context that the function _should_ be called in. It makes me wonder if it would be better to indicate that functionality in the documentation or name of the function, i.e. either rename the function to something like `clearAllSessionDataForReturningUser`, or use the comments for proper documentation outside of the function, e.g. 
```
/// Clears all session data for the user and presents the system logout screen
/// Should only be used when a returning user does not have a persistent ID and cannot reauthenticate
/// - throws: ``PersistentSessionError(.cannotDeleteData)`` in case the user data was not succesfully deleted.
    private func clearAllSessionData() async throws {
```

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -201,43 +202,87 @@ final class PersistentSessionManager: SessionManager {
...
+                analyticsService.logCrash(PersistentSessionError(.noSessionExists,
+                                                                 reason: "secure wallet data deleted"))
+            }
+            try await clearAppForLogin()
+        } catch {
+            throw PersistentSessionError(.cannotDeleteData, originalError: error)
+        }
+    }
+    
+    /// Asserts that a ``persistentID`` is present and accessible for either a returning or a first time user.
```
jamielutton-gds, line 231: Is this correct? I don't think this function validates that `persistentID` is present for a first time user (since it is valid for `persistentID` to be `nil`, and the function will not throw any errors and just return). Is it more accurate to say "Asserts that if there is no ``persistentID`` present or if it cannot be read, then the user must be a first time user. If the user is a returning user without a `persistentID`, all of their data is deleted."

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: It's not clear to me how we do our grouping to be honest. Is it:
* by an access modifier (e.g. all private grouped together)
* logical grouping?
* related functionality?

I don't think there is a single answer but happy to hear what you think. Otherwise, I'd leave it as is for now given we are close to the release.

jamielutton-gds: Happy to leave for now - I would want you / Chionia / Shukri to discuss this together to see what works best for you as well. But I do think for functions that are extremely similar or have the same name but overloaded with different signatures, having them grouped together might be useful. Let's leave though as it's a wider discussion point.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: Agree that assert might be too strong of a word since the function only throws for a returning user. What I am trying to capture is how the affects of this function only take place in the "unexpected" case of `persistentId` being nil.

The way I think of it (and is intended to be read) is that the effects of this function only relate in the case where "``persistentID`` is NOT present" or "Asserts that a ``persistentID`` is present".

Having said that, if you came into it and was confused by the "assert" sounds like the existing name can be improved.

I just don't really have any good suggestions!

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/commits/40854f861d118b812a93e69c5342e80397bd2280

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/commits/40854f861d118b812a93e69c5342e80397bd2280

Tests/UnitTests/Qualification/AppQualifyingServiceTests.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/commits/40854f861d118b812a93e69c5342e80397bd2280

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/914/commits/40854f861d118b812a93e69c5342e80397bd2280

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
jamielutton-gds: Would something like `assertPersistentSessionIsValid` be a suitable name?

qnoid: It sounds like a mouthful to me :) Maybe I misunderstood your comment. I thought you objected to the use of `assert`, not the "session" part.

The `assertPersistentSessionIsValid` seems too specific? What I see in the code is a function called `startAuthSession` which needs to "assert" that it can `performLoginFlow` on the session. How that assertion is evaluated seems like an implementation detail?

PR #924, opened 2026-09-14
fix: DCMAW-21354 a refresh token will be reused, in case of a secure store error.

qnoid:
# [iOS | In case of a secure store error, a refresh token will be reused](https://govukverify.atlassian.net/browse/DCMAW-21354)

* The primary changes proposed by this PR aim to reduce the likelihood of a refresh token reuse.
* Secondary to that, reduce the number of Local Authentication errors as part of resuming a user session.
* Finally, this code makes the codebase safer by encapsulating the refresh token exchange within the `PersistentSessionManager`, increase its cohesion and removes the coupling of the `NetworkingService` to the `RefreshTokenExchangeManager`.

# Changes

There are 3 distinct changes. They are best reviewed in this order:
* [update MockSecureStoreService](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/89a753a9d21b37a8f0adba81d5453c060cab62a2)
* [reduce the likelihood of a refresh token reuse](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/754918e0636d790545ddceb9745f07da698e5fcf)
* [update NetworkingService](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/9cb1ee1ea218c296c3e908d86379270222c7d1c5)

## Updated depencies
* [mobile-ios-secure-store/releases/tag/4.8.0](https://github.com/govuk-one-login/mobile-ios-secure-store/releases/tag/4.8.0)
* * [4.7.0...4.8.0](https://github.com/govuk-one-login/mobile-ios-secure-store/compare/4.7.0...4.8.0)

# Testing
The ticket raised offered proof of how a refresh token will be reused in case of a `SecureStoreError`. While attempting to reproduce the defect in real world scenario, it turned out that the underlying error of `SecureStoreError`, cannot be a `LAError(.systemCancel)`, therefore the test is now written with an underlying error in the  `NSOSStatusErrorDomain`and a code of [errSecInteractionNotAllowed](https://developer.apple.com/documentation/security/errsecinteractionnotallowed). Otherwise, the test remains the same, now called:

* expect_refreshToken_notExchanged_given_keyInteractionNotAllowed_when_resumeSession

Additional test coverage is provided by
* expect_refreshToken_exchanged_when_resumeSession

## Regression Testing

* A returning user can add a credential in their wallet.
* Given the app enters the foreground, when the user session resumes, the app successfully performs a refresh token exchange.
* A new user can perform a login
* * Add a credential in their wallet

https://github.com/user-attachments/assets/f161bd61-d156-47e2-97c1-c2d5fea438cc

# Do not merge

This PR depends on:
* https://github.com/govuk-one-login/mobile-ios-secure-store/pull/81

# Scout 
* Encapsulate encoding/decoding of `StoredTokens` to increase reusability.

# Related
There is a known issue "[when the request to update the token fails with a network error](https://govukverify.atlassian.net/browse/DCMAW-20928)". More specifically, when there is a network error, after the request has been processed by the server but before the response has reached the client.", It's reasonable to assume the number of occurrences to drop, not eliminated, by this change.

# Observability
Once the changes in this PR have made it to the next release, would be good to keep an eye on any:
* `SecureStoreError` as they relate to local authentication; a number of them are identified [in the ticket](https://govukverify.atlassian.net/browse/DCMAW-21354).
* Reports of a refresh token reuse and [account intervention](https://console.firebase.google.com/project/mobile-one-login-prod/crashlytics/app/ios:uk.gov.onelogin/issues/4187f3ad1004064827e87ce508693418?time=90d&subIssues=61087923dfedb08b9e7fb3bfa489ea04&sessionEventKey=47b78e70934d41f7b643968078c8d19f_2264934635700718345).

I would expect to see a reduction in the number of occurrences.

# Checklist


Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -383,11 +386,28 @@ final class PersistentSessionManager: SessionManager {
         }
     }
     
-    func saveLoginTokens(
+    @MainActor
+    func refreshTokens(idToken: String, refreshToken: String) async throws {
+        let encryptor = try accessControlEncryptedStore.encryptor()
```
jamielutton-gds, line 390: How does updating this code using the 'two-step encryption' approach reduce the likelihood of a refresh token being reused?

> - Use a two-step encryption so that a refresh token request happens *after* ensuring that the key is accessible.
	This is done in an attempt to reduce the number of errors that can occur during saving (i.e. encrypting)
	the stored tokens thus leading to a now refresh token being reused, causing an account an intervention
	next time the refresh token is exchange is performed.

If the _public key needed to encrypt_ isn't accessible, wouldn't that imply that the _private key needed to decrypt_ the refresh token also isn't accessible, [at the point where self.storeKeyService.fetch is called](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes#diff-d9e901b931fb93a0c3733e4284970bf1a470ddf02ad8e7985adf02e7cbd03299R357), hence, the app would never get as far as performing the refresh token exchange with STS, and hence we would never reach this step of trying to save the data?

Or is there a scenario I'm missing where it would be possible to access the private key to decrypt the data, but not possible to access to public key to encrypt it?

qnoid: > Or is there a scenario I'm missing where it would be possible to access the private key to decrypt the data, but not possible to access to public key to encrypt it?

In theory, yes. Not sure how likely it is in practice. However improbable, it's still possible and error handling should be robust. Especially since we currently have no way of retrying a refresh token exchange. The risk is leaving the app in an invalid state.

It's more than whether the key is present or not. [The Keychain can throw any number of errors](https://developer.apple.com/documentation/security/security-framework-result-codes#Keychain-Result-Codes), some of them transient. That is to say, just because the key was successfully retrieved earlier, should not be used as a strong guarantee the next retrieval will succeed.

There are all sorts of underlying `NSOSStatusErrorDomain` errors that can occur [when fetching the private key](https://github.com/govuk-one-login/mobile-ios-secure-store/blob/main/Sources/SecureStore/KeyManagerService.swift#L142).

For what is worth, since we started capturing the [underlying error](https://github.com/govuk-one-login/mobile-ios-secure-store/pull/70) for the `KeyManagerService`, in our logs, I can only see "[The specified item could not be found in the keychain.](https://console.firebase.google.com/project/mobile-one-login-prod/crashlytics/app/ios:uk.gov.onelogin/issues/0694bc52aef714256b5fc25490f47e3f?time=90d&customKeys=kind%E2%87%92cantRetrieveKey&sessionEventKey=a78486fa50e5471191fed71d85d1bc0c_2263869296528118390)" which is a permanent error. With that log alone, the circumstances that lead to that are less clear.

Sources/Utilities/Storage/Session/SessionManager.swift
```diff
@@ -36,16 +36,26 @@ protocol SessionManager: AnyObject, UserProvider {
...
-        refreshToken: String?,
-        accessToken: String?,
-        accessTokenExpiry: Date?
-    ) throws
-    
     /// Resumes an existing session by restoring tokens from on-device storage
+    @MainActor
     func resumeSession() async throws
+
+    /// Perform a refresh token exchange for the given `idToken` using the given `refreshToken`.
```
jamielutton-gds: The refresh token doesn't refresh the `idToken`, it refreshes the access token. 

It looks like the only reason that `refreshTokens` needs to accept the `idToken` as a parameter is because the response from STS _when the refresh token is exchanged_ does not contain an `id_token`*, and since the access, refresh and ID tokens are saved in the same blob of data, the `refreshTokens` function needs access to the previously saved `idToken` so that when it saves the updated access and refresh tokens, it ensures that the last issued `idToken` is still saved alongside them.

So I think there are two things to note here:
(1) This comment should be updated to make it clear that the ID token is not the token that is refreshed
(2) It's maybe a tad confusing that a function called `refreshTokens` accepts an `idToken` when it is not needed to perform the refresh, but only to ensure consistency in the data stored, but that itself is a side-effect of the fact that it's not possible to only update stored tokens to save the new access and refresh token, and that the `idToken` is currently coupled to them. It's maybe not a problem that needs to be addressed on this PR since the aim of this PR is to reduce likelihood of refresh token reuse, but just flagging that this could potentially be improved in future to make the function signature easier to understand by somehow removing the need for an `idToken` to be passed in but ensuring that the previous one still persists in the secure store. 

[*]: The response from STS for a refresh token exchange looks like this:
```
{
  "access_token": "<some access token>",
  "refresh_token": "<some refresh token>",
  "expires_in": 3600,
  "token_type": "Bearer"
}
```

Sources/Utilities/Storage/Session/SessionManager.swift
```diff
@@ -36,16 +36,26 @@ protocol SessionManager: AnyObject, UserProvider {
...
+    ///
+    /// Upon a succesful refresh token exchange, the login tokens will be saved for the given token id.
+    ///
+    /// - Parameters:
+    ///     - idToken: the associated token id for which to perform the refresh token exchange
+    ///     - refreshToken: an existing, valid refresh token to be used for the refresh token exchange
+    ///  - throws: ``SecureStoreError(.cantRetrieveKey)`` in case there was an error retrieving the
```
jamielutton-gds: Looks like something more was meant to be said here

Sources/Utilities/Storage/Session/SessionManager.swift
```diff
@@ -36,16 +36,26 @@ protocol SessionManager: AnyObject, UserProvider {
...
+    ///  - throws: ``ServerError`` in case of a 400 response
+    ///  - throws: ``RefreshTokenExchangeError.noInternet`` in case the network connection was either never
+    ///      established, timed out or lost during the refresh token exchange.
+    ///  - throws: a
```
jamielutton-gds: Was something else meant to be said here?

Sources/Utilities/Storage/Session/SessionManager.swift
qnoid: Thanks for breaking it down Jamie! I opted not to make a bigger refactor to be honest. Definitely one for a future improvement.

In regards to the comment, the way I read it is how the refresh token exchange is "associated" with the given `idToken` (a refresh token currently stored alongside the `idToken` like you said) but the refresh is done "using" the given refresh token.

Happy to hear a suggestion if you think the documentation can be improved. Writing documentation is hard!

Sources/Utilities/Storage/Session/SessionManager.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/c19c4747b06f035138e95f9beb08b6f27b568656

Sources/Utilities/Storage/Session/SessionManager.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/c19c4747b06f035138e95f9beb08b6f27b568656

Sources/Utilities/Storage/Session/SessionManager.swift
jamielutton-gds: I don't think it's really right to say that a refresh token is "associated" with an id_token, it's more closely related to the access_token. It might be more accurate to say something like:
```
/// Perform a refresh token exchange for the current session using the given `refreshToken`.
/// Upon a successful refresh token exchange, the refreshed `accessToken` and `refreshToken` for this session will be saved with the `idToken`.
```
I admit it is difficult to bring in the `idToken` here in an appropriate way, but the key thing is that the tokens are all linked to a session, but I don't think it's really right to say that we are doing a refresh "for" the `idToken` because that would imply that the `idToken` is the token that is being refreshed, which is not the case. 

I don't want to get too hung up on the documentation here - happy if you want to try and find a middle ground between what you have and what I've suggested if you like, but also, I think it would be much easier to not need to refer to the idToken at all here, but that is a ~symptom~ side effect of the other problem I mentioned above, which is not within the scope of this PR to fix!

qnoid: Thank you Jamie! That clarified a lot in my head of how the whole thing works.
Appreciate the suggested documentation:

https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/36d0742e41ad42ffec1c426ba6c430f49ffab27e

jamielutton-gds (approved):
I have not had enough time to review the tests for this in detail, but I am giving an approval to show that I think the source code changes look fine. Would appreciate that someone else reviews and making sure that they've given a good review of the unit test changes.

qnoid:
> I have not had enough time to review the tests for this in detail, but I am giving an approval to show that I think the source code changes look fine. Would appreciate that someone else reviews and making sure that they've given a good review of the unit test changes.

Thank you for clarifying Jamie! That is helpful context.

OneLogin.xcodeproj/project.pbxproj
```diff
@@ -5587,8 +5587,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/govuk-one-login/mobile-ios-secure-store";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 4.5.0;
+				branch = "fix/DCMAW-21354";
+				kind = branch;
```
shukriahmed-dd: leaving a comment here so it can be resolved once updated :)

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -382,11 +385,28 @@ final class PersistentSessionManager: SessionManager {
         }
     }
     
-    func saveLoginTokens(
+    @MainActor
+    func refreshTokens(idToken: String, refreshToken: String) async throws {
```
shukriahmed-dd: nit: rename to `updateRefreshToken` ?

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -382,11 +385,28 @@ final class PersistentSessionManager: SessionManager {
...
+            idToken: idToken,
+            refreshToken: exchangeTokenResponse.refreshToken,
+            accessToken: exchangeTokenResponse.accessToken,
+            accessTokenExpiry: exchangeTokenResponse.expiryDate,
+            saveLoginTokens: { tokens in
+                try self.storeKeyService.save(using: encryptor, tokens: tokens)
+            })
+    }
+    
+    private func update(
```
shukriahmed-dd: nit: could the name here stay as `saveLoginTokens`, with the parameter on line 409 being renamed to `storeLoginTokens`. I think these names are more descriptive

Sources/Utilities/Storage/SecureTokenStore.swift
```diff
@@ -14,15 +14,33 @@ public struct StoredTokens: Codable {
...
+            throw StoredTokenError.unableToDecodeTokens
+        }
+        self = try JSONDecoder().decode(Self.self, from: tokensAsData)
+    }
+    
+    public func base64EncodedJSON() throws -> String {
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = .sortedKeys
+        let data = try encoder.encode(self)
+        return data.base64EncodedString()
```
shukriahmed-dd, line 36: Was this extension created to improve readability? I find the previous style easier to read. Looks like this function `base64EncodedJSON` isnt being used elsewhere as well

mobile-wallet-ios
shukriahmed-dd, line 1: Looks like wallet has been updated, can this be reverted please

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: 🤔 Technically it's [an update](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes#diff-d9e901b931fb93a0c3733e4284970bf1a470ddf02ad8e7985adf02e7cbd03299R428), that [includes a save](https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes#diff-d9e901b931fb93a0c3733e4284970bf1a470ddf02ad8e7985adf02e7cbd03299R427), no?

Even though the function has other side effects like deleting the refreshTokenExpiry.

Sources/Utilities/Storage/SecureTokenStore.swift
qnoid: It was not done to improve readability. It was done for encapsulation and code safety.
As the code stands, it both serialises and deserialises a `StoredTokens` instance to and from `base64EncodedJSON`.

Neither requires a relationship to `SecureTokenStore` to perform. Equally, any type (like the `SecureTokenStore` here) that needs to serialise an instance of `StoredTokens`, can use the:

*  _**base64EncodedJSON**_ function to serialise
* use an initialiser _init(**base64EncodedJSON**: String)_ to deserialise

It's just a good convention to follow and look for when you want the return value of one function to be the input to another.

mobile-wallet-ios
qnoid: Will do before the merge. I wanted to ensure Wallet compiles using the `fix/DCMAW-21354` branch.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/032a3bf53feb1dca5ac6f3fbac6ebd556200d35b

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -382,11 +385,28 @@ final class PersistentSessionManager: SessionManager {
...
         idToken: String?,
         refreshToken: String?,
         accessToken: String?,
-        accessTokenExpiry: Date?
+        accessTokenExpiry: Date?,
+        saveLoginTokens: (StoredTokens) throws -> Void
```
chionia-gds: shoudn't this be an auto closure

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
```diff
@@ -403,9 +423,8 @@ final class PersistentSessionManager: SessionManager {
             accessToken: accessToken,
             accessTokenExpiry: accessTokenExpiry
         )
-        
-        try storeKeyService.save(tokens: tokens)
-        
+
+        try saveLoginTokens(tokens)
```
chionia-gds: why have we replaced the save with this function? we call it every time anyway

qnoid: Because the `save` function called on the `TokenStore` varies between 
* [saveAuthSession](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/032a3bf53feb1dca5ac6f3fbac6ebd556200d35b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L337), which matches the previous implementation
* [updateRefreshToken](https://github.com/govuk-one-login/mobile-ios-one-login-app/blob/032a3bf53feb1dca5ac6f3fbac6ebd556200d35b/Sources/Utilities/Storage/Session/PersistentSessionManager.swift#L400), which uses an encryptor to save.

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: I am not sure how this can be an @autoclosure. Can you please elaborate?

chionia-gds: yeah you're right, can't be with an argument passed, sorry

OneLogin.xcodeproj/project.pbxproj
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/e4a3979900262d7c125bdf8c4cbe4cf3d290b6e0

mobile-wallet-ios
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/changes/e4a3979900262d7c125bdf8c4cbe4cf3d290b6e0

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: No worries! I might have learned something new :)

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
shukriahmed-dd: Yes , but i think the name isn't descriptive enough, how about `updateLoginTokens` ?

Sources/Utilities/Storage/AccessControlEncryptedSecureStoreMigrator.swift
```diff
@@ -51,6 +51,16 @@ final class AccessControlEncryptedSecureStoreMigrator: SecureStorable, SessionBo
...
     }
     
+    func encryptor() throws -> Encryptor {
+        return try self.v13AccessControlEncryptedSecureStore.encryptor()
+    }
+
+    func save(using encryptor: Encryptor, item: String, itemName: String) throws {
+        try self.v13AccessControlEncryptedSecureStore.save(using: encryptor, item: item, itemName: itemName)
+
+        hasMigrated = true
```
shukriahmed-dd, line 61: How can we confirm that `readItem` is able to still read from a save that uses an encryptor. I see on the PR theres a regression test adding a credential. To confirm we can still read - would closing and reopening the app (so the credential is displayed) check this ?

Sources/Utilities/Storage/SecureTokenStore.swift
shukriahmed-dd: thanks for explaining

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: Sounds like a good suggestion to me. Do you think we should go further and name it:

```suggestion
    private func updateLoginTokens(
        idToken: String?,
        refreshToken: String?,
        accessToken: String?,
        accessTokenExpiry: Date?,
        save: (StoredTokens) throws -> Void
) throws
```

Sources/Utilities/Storage/SecureTokenStore.swift
qnoid: Thanks for being curious :)

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
shukriahmed-dd: yes thats great, thanks

Sources/Utilities/Storage/AccessControlEncryptedSecureStoreMigrator.swift
qnoid: The simple answer to this is that since the `encryptor` is returned from the same `SecureStore` and by extension its `KeyManagerService` instance it has a relationship with, then it's bound to be able to read it. Don't you think that?

What your question got me thinking is what should the relationship be between the `save(using:item:itemName)` and the `saveItem(item:itemName)` function and by association the `readItem(itemName:)` if any. 🤔

Sources/Utilities/Storage/Session/PersistentSessionManager.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/commits/069da4de5f3568f14ef2cc8d58b3530a06a0ed77

Sources/Utilities/Storage/AccessControlEncryptedSecureStoreMigrator.swift
qnoid: https://github.com/govuk-one-login/mobile-ios-one-login-app/pull/924/commits/bda33bf701110505eecf4c83e1bbd469da7e3891

qnoid: Assuming I am reading the `AccessControlEncryptedSecureStoreMigrator` right, the only postcondition as it relates to a save function (encryptor or no encryptor) is to set the `hasMigrated` to `true`. I think that's all in terms of its public contract. @shukriahmed-dd what do you think?

chionia-gds (approved):

Sources/Utilities/Storage/AccessControlEncryptedSecureStoreMigrator.swift
shukriahmed-dd: Yes I agree, and the test gives us extra confidence :)

shukriahmed-dd (approved):

